PR #32, opened 2014-12-15
Reviewer Mike - Multiple namespaces

bossmc:


cdivdata.cpp
```diff
@@ -51,8 +51,7 @@ BareStatHandler cdiv_not_registered_handler(cdiv_not_registered_oid, &tree);
 BareStatHandler cdiv_no_answer_handler(cdiv_no_answer_oid, &tree);
 BareStatHandler cdiv_not_reachable_handler(cdiv_not_reachable_oid, &tree);
 
-NodeData cdiv_node_data("cdiv_handler",
-                        "6666",
+NodeData cdiv_node_data("sprout",
```
mike-evans: Isn't this baking in an assumption that the CDIV AS will always run collocated with Sprout?  We still have the option of running ASs standalone.

mementoasdata.cpp
```diff
@@ -49,8 +49,7 @@ BareStatHandler memento_not_recorded_overload_handler(memento_not_recorded_overl
 AccumulatedWithCountStatHandler memento_cassandra_read_latency_handler(memento_cassandra_read_latency_oid, &tree);
 AccumulatedWithCountStatHandler memento_cassandra_write_latency_handler(memento_cassandra_write_latency_oid, &tree);
 
-NodeData memento_as_node_data("memento_as_handler",
-                              "6666",
+NodeData memento_as_node_data("sprout",
```
mike-evans: As per CDIV AS comment above.

PR #40, opened 2015-02-06
Use ipc instead of tcp for 0MQ alarm interface

dtheobald:
These changes are required for alarms to work in a separated mgmt/signalling network environment. The changes are similar in nature to those needed for statistics in that environment and have been tested in a multiple network environment in system test.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -171,6 +171,7 @@ test: run_test coverage vg coverage-check vg-check
 # Ignore failure here; it will be detected by Jenkins.
 .PHONY: run_test
 run_test: build_test | $(TEST_OUT_DIR)
+	mkdir -p /var/run/clearwater
 	rm -f $(TEST_XML)
 	rm -f $(OBJ_DIR_TEST)/*.gcda
 	$(TARGET_BIN_TEST) $(EXTRA_TEST_ARGS) --gtest_output=xml:$(TEST_XML)
--- a/Makefile
+++ b/Makefile
@@ -171,7 +171,7 @@ test: run_test coverage vg coverage-check vg-check
 # Ignore failure here; it will be detected by Jenkins.
 .PHONY: run_test
 run_test: build_test | $(TEST_OUT_DIR)
-	mkdir -p /var/run/clearwater
+	(mkdir -p /var/run/clearwater; if [ $$? -ne 0 ]; then ( printf "\nFor tests to run, as root do the following:\n  $$ mkdir -p /var/run/clearwater\n  $$ chmod -R o+wr /var/run/clearwater\n" ); exit 1; fi)
 	rm -f $(TEST_XML)
 	rm -f $(OBJ_DIR_TEST)/*.gcda
 	$(TARGET_BIN_TEST) $(EXTRA_TEST_ARGS) --gtest_output=xml:$(TEST_XML)
--- a/alarm_req_listener.cpp
+++ b/alarm_req_listener.cpp
@@ -36,6 +36,9 @@
 #include <zmq.h>
 
 #include <sstream>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <signal.h>
 
 #include <net-snmp/net-snmp-config.h>
 #include <net-snmp/net-snmp-includes.h>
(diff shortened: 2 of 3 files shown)


alarm_req_listener.cpp
```diff
@@ -123,11 +126,25 @@ bool AlarmReqListener::zmq_init_sck()
     return false;
   }
 
-  std::stringstream ss;
-  ss << "tcp://127.0.0.1:" << ZMQ_PORT; 
+  std::string sf = std::string("/var/run/clearwater/alarms");
+  std::string ss = std::string("ipc://" + sf);
```
AlexHockey: The original variable was called `ss` because it was a `stringstream`. This is no longer the case. I would also prefer slightly more descriptive names, maybe `sck_dir` and `sck_url`?

alarm_req_listener.cpp
```diff
@@ -138,6 +155,12 @@ bool AlarmReqListener::zmq_init_sck()
     return false;
   }
 
+  rc=chmod(sf.c_str(), 0777);
+  if (rc == -1)
+  {
+      snmp_log(LOG_ERR, "chmod(%s, 0777) failed: %s", sf.c_str(), strerror(errno));
```
AlexHockey: Indentation looks wrong (should be 2 spaces but looks like 4)

alarm_req_listener.cpp
```diff
@@ -123,11 +126,25 @@ bool AlarmReqListener::zmq_init_sck()
...
 
   int rc;
-  while (((rc = zmq_bind(_sck, ss.str().c_str())) == -1) && (errno == EINTR))
+  rc=remove(sf.c_str());
+  if (rc == -1)
+  {
+    if (errno != ENOENT)
+    {
+      snmp_log(LOG_ERR, "remove(%s) failed: %s - killing myself", sf.c_str(), strerror(errno));
+      kill(getpid(), SIGKILL);
```
AlexHockey: `KILL`ing the current process feels extreme. Why not just use `exit()`?

alarm_req_listener.cpp
```diff
@@ -123,11 +126,25 @@ bool AlarmReqListener::zmq_init_sck()
...
+    }
+    else
+    {
+      snmp_log(LOG_ERR, "remove(%s) failed: %s", sf.c_str(), strerror(errno));
+    }
+  }
+  while (((rc = zmq_bind(_sck, ss.c_str())) == -1) && (errno == EINTR))
```
AlexHockey: My understanding is that `zmq_bind` will unlink the socket for you, so the entire preceding code block is unnecessarily.

alarm_req_listener.cpp
```diff
@@ -123,11 +126,25 @@ bool AlarmReqListener::zmq_init_sck()
...
+    if (errno != ENOENT)
+    {
+      snmp_log(LOG_ERR, "remove(%s) failed: %s - killing myself", sf.c_str(), strerror(errno));
+      kill(getpid(), SIGKILL);
+    }
+    else
+    {
+      snmp_log(LOG_ERR, "remove(%s) failed: %s", sf.c_str(), strerror(errno));
+    }
+  }
```
AlexHockey: Should have a blank line here to visually separate the `if` block from the `while` below.

Makefile
```diff
@@ -171,6 +171,7 @@ test: run_test coverage vg coverage-check vg-check
 # Ignore failure here; it will be detected by Jenkins.
 .PHONY: run_test
 run_test: build_test | $(TEST_OUT_DIR)
+	mkdir -p /var/run/clearwater
```
AlexHockey: On all the machines I've looked at `/var/run` is `drwxr-xr-x`` and owned by `root:root`. This means if I run my unit tests as a normal (non-root) I can't create this directory. It would be better to try writing to it and if unsuccessful, print a message telling users they need to create it (possibly as a root)

alarm_req_listener.cpp
dtheobald: I don't want any (SNMPD) exit handler to run.

alarm_req_listener.cpp
dtheobald: I was just following Andy's lead in cpp-common/src/zmq_lvc.cpp

Makefile
dtheobald: How do the stats unit tests run? They should have the same problem, so I should probably do what they do, but I'm not sure where they are located.

AlexHockey: I think we only have UTs for alarms, not stats.

alarm_req_listener.cpp
AlexHockey: Bypassing the snmpd exit handlers is a bold move. Can you provide some more detail on why this is required?

alarm_req_listener.cpp
AlexHockey: Ok. I suppose there is no harm in deleting the socket twice, so fine.

AlexHockey:
Doug, I'm assigning this PR back to you while you make the markup I suggested.

eleanor-merry:
@AlexHockey What's the status on this PR?

AlexHockey:
Blocked waiting for @dtheobald to address code review feedback in https://github.com/Metaswitch/clearwater-infrastructure/pull/134

eleanor-merry:
Merged in https://github.com/Metaswitch/clearwater-snmp-handlers/commit/4791afda4204d465f8b41a29c4fe6e08383ab38c

PR #44, opened 2015-03-04
[Reviewer: Andy] Add Astaire statistic MIBs

mirw:
Andy,

Please can you review my MIB definition for the Astaire statistics?

Thanks,

Matt


PROJECT-CLEARWATER-MIB
```diff
@@ -2447,4 +2450,237 @@ mementoGroup OBJECT-GROUP
...
+                 resynchronization operation."
+    ::= { astaire 1 }
+
+astaireBucketsResynchronized OBJECT-TYPE
+    SYNTAX      Unsigned32
+    MAX-ACCESS  read-only
+    STATUS      current
+    DESCRIPTION "Number of buckets already resynchronized during the current
+                 resynchronization operation."
+    ::= { astaire 2 }
```
bossmc: What's the value of this statistic if there are no ongoing syncs?

bossmc: And throughout.

PROJECT-CLEARWATER-MIB
```diff
@@ -2447,4 +2450,237 @@ mementoGroup OBJECT-GROUP
...
+                 connection for the current resynchronization operation."
+    ::= { astaireConnectionTableEntry 3 }
+
+astaireConnectionBucketsResynchronized OBJECT-TYPE
+    SYNTAX      Unsigned32
+    MAX-ACCESS  read-only
+    STATUS      current
+    DESCRIPTION "Number of buckets already resynchronized over this connection
+                 during the current resynchronization operation."
+    ::= { astaireConnectionTableEntry 4 }
```
bossmc: Since we don't know when a bucket finishes resynch, this will always either equal the previous stat or 0.  I think we're ok with this, but I'm just checking.

PROJECT-CLEARWATER-MIB
```diff
@@ -2447,4 +2450,237 @@ mementoGroup OBJECT-GROUP
...
+                 bucket."
+    INDEX { astaireConnectionBucketId }
+    ::= { astaireConnectionBucketTable 1 }
+
+AstaireConnectionBucketTableEntry ::= SEQUENCE
+{
+  astaireConnectionBucketId                     Unsigned32,
+  astaireConnectionBucketEntriesResynchronized  Unsigned32,
+  astaireConnectionBucketDataResynchronized     Unsigned32
+}
```
bossmc: How long do these stats last?  The connection disappears after any related resynch operations complete (and might be recreated to handle reads from backups).

bossmc:
No technical queries, some spec-behaviour questions though.

PROJECT-CLEARWATER-MIB
mirw: Agreed.  I think this is OK.

PROJECT-CLEARWATER-MIB
mirw: These objects represent active connections and are deleted as soon as the connection terminates.  I've clarified the comments.

PROJECT-CLEARWATER-MIB
mirw: If there's no ongoing synchronization, there is no resynchronization operation, and these return 0.  I've updated the definition.

mirw:
@bossmc, please can you now review the implementation?  I've tested live, using both snmpwalk and snmptable (to confirm that tables are rendered correctly).

astairedata.cpp
```diff
@@ -0,0 +1,189 @@
...
+  }
+};
+
+OID astaire_oid = OID("1.2.826.0.1.1578918.9.9");
+
+AstaireGlobalStatHandler global_handler(astaire_oid, &tree);
+AstaireConnectionStatHandler connection_handler(astaire_oid, &tree);
+
+NodeData astaire_node_data("astaire",
+                           OID("1.2.826.0.1.1578918.9.9"),
```
bossmc: Isn't this `astaire_oid`?

bossmc:
Looks good.

PR #45, opened 2015-03-10
[Reviewer: Ellie] Fixes for 14.04

rkd-msw:
I hate threads so much.

Between net-snmp 5.4 (in 12.04) and 5.7 (in 14.04), snmpd's startup order has changed - it used to fork then load all the .so plugins, it now loads all the .so plugins then forks. This means that the process that creates the stats thread immediately exits, so snmpd never gets any stats over 0MQ.

This change just lazily-loads the thread, when we receive the first request - this ensures that it starts in the surviving child process, not the dead parent process.


custom_handler.cpp
```diff
@@ -93,6 +95,13 @@ int clearwater_handler(netsnmp_mib_handler* handler,
   netsnmp_request_info* request;
   netsnmp_variable_list* var;
 
+  if (thread_created.load() != true)
+  {
+    thread_created.store(true);
+    pthread_t zmq_thread;
+    pthread_create(&zmq_thread, NULL, start_stats, global_node_data);
+  }
```
bossmc: Race condition.

eleanor-merry:
No further comments from me

PR #54, opened 2015-04-09
[Reviewer: Matt] Add chronos stats

eleanor-merry:
Matt, can you please review these changes to add Chronos stats


mirw:
Looks good.  Only comment is that these statistics are quite bare-bones.  Are we exposing as much information as we could/should?  What statistics are we *not* exposing?

PR #55, opened 2015-04-22
[Reviewer: Matt] Update stats

eleanor-merry:
Matt, can you review the updated snmp handlers with the new Diameter routing stats


mirw:
Why aren't these part of the same table, or even additional fields on the incoming requests table?

PR #61, opened 2015-06-22
[Reviewer: Matt] Scope changes

rkd-msw:
Corresponding changes to https://github.com/Metaswitch/cpp-common/pull/311.

* My new code exposes three scopes for time-based tables (prev 5 seconds, current 5 minutes, prev 5 minutes) so I could make sure my code was sufficiently flexible. I've now updated this to match.
* I also took the opportunity to correct `scopeCurrent5SecondPeriod` to `scopePrevious5SecondPeriod` (which I think we agreed was more accurate even for existing stats).
* I noticed that there were some references to "hss" which should have been "homestead", so I've fixed those up.

### DIFF
--- a/PROJECT-CLEARWATER-MIB
+++ b/PROJECT-CLEARWATER-MIB
@@ -586,14 +586,14 @@ sproutHomerLatencyCount OBJECT-TYPE
     DESCRIPTION "The total requests over the period."
     ::= { sproutHomerLatencyEntry 6 }
 
-homestead OBJECT IDENTIFIER ::= { sprout 3 }
+hss OBJECT IDENTIFIER ::= { sprout 3 }
 
 sproutConnectedHomesteadsTable OBJECT-TYPE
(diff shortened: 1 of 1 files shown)


PROJECT-CLEARWATER-MIB
```diff
@@ -571,14 +586,14 @@ sproutHomerLatencyCount OBJECT-TYPE
     DESCRIPTION "The total requests over the period."
     ::= { sproutHomerLatencyEntry 6 }
 
-hss OBJECT IDENTIFIER ::= { sprout 3 }
+homestead OBJECT IDENTIFIER ::= { sprout 3 }
```
mirw: Doesn't this conflict with the other homestead definition at line 71?

mirw:
Generally looks sensible.  One comment.

PROJECT-CLEARWATER-MIB
rkd-msw: Yes - I'll revert this for now and think about how to name this better.

PR #63, opened 2015-06-26
[Reviewer: Andy] Set a ZMQ_LINGER option so we quit when told to

rkd-msw:
Fixes the hang we're seeing on the repo server.


bossmc:
Sure.

PR #66, opened 2015-07-09
Review AMC: Don't report the same value if SNMP GET performed for multiple OIDs

steelyvoid:
Bug here is that if you issue an SNMP GET to a node specifying mulitple OIDs, you get the value of the first reported for all of them.

Fix tested live using snmpget with mulitple OIDs


bossmc:
Haha, good spot.

PR #71, opened 2015-07-19
[Reviewer: Ellie] Allow snmp_ip to be a comma-separated list

### LINKED ISSUE #70: We no longer support multiple IP addresses in snmp_ip
We used to allow multiple IP addresses in snmp_ip; we now don't. Using a comma-separated string in snmp_ip will stop any statistics from being sent. 

Marking as high priority there's no workaround to allow multiple IP addresses, and we rely on statistics for diagnosing issues/alarming.

rkd-msw:
Tested live, by setting it to two IPs and confirming that alarms were sent to both. Fixes #70.


eleanor-merry:
Looks good

PR #72, opened 2015-07-27
[Reviewer: Rob] Parse from multiple files

eleanor-merry:
Rob, can you please review this change to have the alarms agent parse its alarms from JSON files, not cpp-common. This also holds the UTs for the common alarm parsing code.

### DIFF
--- a/alarm_table_defs.cpp
+++ b/alarm_table_defs.cpp
@@ -94,6 +94,7 @@ bool AlarmTableDefs::initialize(std::string& path)
   else
   {
     snmp_log(LOG_ERR, "Unable to open directory at %s", path.c_str());
+    rc = false;
   }
 
   return rc;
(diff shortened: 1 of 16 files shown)


alarm_table_defs.cpp
```diff
@@ -69,47 +67,61 @@ bool AlarmTableDefKey::operator<(const AlarmTableDefKey& rhs) const
...
+
+      dir_it++;
+    }
+  }
+  else
   {
-    std::vector<AlarmDef::AlarmDefinition> local_alarms;
-    init_ok = parse_local_alarms_from_file(path, local_alarms);
-    init_ok &= populate_map(local_alarms, dup_check);
+    snmp_log(LOG_ERR, "Unable to open directory at %s", path.c_str());
```
rkd-msw: Should we set `rc = false` here?

rkd-msw:
Looks good. One comment about behaviour when the directory doesn't exist (or is empty?) - I think we should repeatedly restart in that case.

PR #73, opened 2015-07-28
[reviewer: Rob] Non register auth table

alango:


rkd-msw:
Looks good, no comments.

PR #79, opened 2015-08-06
[Reviewer: Rob] SNMP fixes

### LINKED ISSUE #76: alarms_agent doesn't log to syslog
I couldn't see any logs generated by the alarm_agent (specifically alarm_req_listener.cpp) and noticed that the snmp subagent start up (alarms_agent.cpp) was not enabling syslog, so I enabled it as follows:
```
@ -75,11 +75,11 @@ int main (int argc, char **argv)
         abort ();
       }
   }
 
   // Log SNMP library output to syslog
-  snmp_enable_calllog();
+  snmp_enable_syslog_ident("alarms_agent", LOG_DAEMON);
 
   // Set ourselves up as a subagent
   netsnmp_ds_set_boolean(NETSNMP_DS_APPLICATION_ID, NETSNMP_DS_AGENT_ROLE, 1);
   init_agent("clearwater-alarms");
 ```

### LINKED ISSUE #77: alarm_req_listener is not handling zmq_bind error
When alarm_req_listener tries to zmq_bind to /var/run/clearwater/alarms and it fails, no action is taken so there are conditions underwhich alarms don't work.

I fixed this by exiting the alarms_agent upon encountering such an error as follows:
```
@@ -158,9 +158,14 @@ bool AlarmReqListener::zmq_init_sck()
 
   if (rc == -1)
   {
-    snmp_log(LOG_ERR, "zmq_bind failed: %s", zmq_strerror(errno));
+    // LCOV_EXCL_START
+    snmp_log(LOG_ERR, "zmq_bind(.., %s) failed: %s - killing myself", sck_url.c_str(), zmq_strerror(errno));
+    kill(getpid(), SIGKILL);
+    // LCOV_EXCL_STOP
     return false;
   }
+  
+  snmp_log(LOG_INFO, "zmq_bind(.., %s) succeeded", sck_url.c_str());
 
   rc=chmod(sck_file.c_str(), 0777);
   if (rc == -1)
```
I resorted to killing because past experience with net-snmp hanging on exit has left me weary of allowing any exit handlers to run.

### LINKED ISSUE #78: alarms_agent.cpp is sending traps to the wrong port
Traps should be sent to port 162:
```
@@ -88,7 +88,7 @@ int main (int argc, char **argv)
        ii != trap_ips.end();
        ii++)
   {
-    create_trap_session(const_cast<char*>(ii->c_str()), 161, community,
+    create_trap_session(const_cast<char*>(ii->c_str()), 162, community,
                         SNMP_VERSION_2c, SNMP_MSG_INFORM);  
   }
```

### PR DESCRIPTION
**Pull request by eleanor-merry**
Rob, can you review these fixes for the alarms agent. The main fix is that we didn't check the output of running start on the AlarmReqListener, so we could run a broken process

Tested live

Fixes #78, fixes #77, fixes #76

### DIFF
--- a/alarm_req_listener.cpp
+++ b/alarm_req_listener.cpp
@@ -161,6 +161,9 @@ bool AlarmReqListener::zmq_init_sck()
   if (rc == -1)
   {
     snmp_log(LOG_ERR, "chmod(%s, 0777) failed: %s", sck_file.c_str(), strerror(errno));
+    // We don't return false in UT as the chmod always fails (because the
+    // previous zmq calls are mocked out the socket file isn't created so can't
+    // be chmodded).
 #ifndef UNIT_TEST
     return false;
 #endif
--- a/alarm_req_listener.cpp
+++ b/alarm_req_listener.cpp
@@ -75,12 +75,9 @@ bool AlarmReqListener::start()
   if (rc != 0)
   {
     // LCOV_EXCL_START - No mock for pthread_create
-
     snmp_log(LOG_ERR, "error creating listener thread: %s", strerror(rc));
-
     zmq_clean_ctx();
     return false;
-
     // LCOV_EXCL_STOP
   }
 
@@ -120,6 +117,7 @@ bool AlarmReqListener::zmq_init_ctx()
 bool AlarmReqListener::zmq_init_sck()
 {
   _sck = zmq_socket(_ctx, ZMQ_REP);
+
   if (_sck == NULL)
   {
     snmp_log(LOG_ERR, "zmq_socket failed: %s", zmq_strerror(errno));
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: alarm_req_listener.cpp [hunk 1]
```diff
@@ -166,6 +161,9 @@ bool AlarmReqListener::zmq_init_sck()
   if (rc == -1)
   {
     snmp_log(LOG_ERR, "chmod(%s, 0777) failed: %s", sck_file.c_str(), strerror(errno));
+#ifndef UNIT_TEST
+    return false;
```
rkd-msw: Why not return false in UT?

alarms_agent.cpp
```diff
@@ -78,6 +78,7 @@ int main (int argc, char **argv)
 
   // Log SNMP library output to syslog
```
rkd-msw: Should we be logging to syslog, or should we be creating /var/log/clearwater-alarms/ and logging there?

rkd-msw:
Looks good. One question, one WIBNI.

alarms_agent.cpp
eleanor-merry: Perhaps, but I'm not going to do so in this PR

Issue #92, opened 2015-10-06
MG-SOFT MIB Browser cannot compile the release-82 PROJECT-CLEARWATER-MIB.


plfrancis:
The following errors and warnings were generated when attempting to compile the release-82 MIB. Both errors should be fixed and the warnings investigated.

```
C:\MIBS\PROJEC~1.MIB(4819) : Warning 30 : No size/range or named values are allowed on SEQUENCE "AstaireConnectionBucketTableEntry" for syntax of "astaireConnectionBucketInetAddr"
C:\MIBS\PROJEC~1.MIB(4893) : Warning 30 : No size/range or named values are allowed on SEQUENCE "AstaireConnectionBucketBandwidthTableEntry" for syntax of "astaireConnectionBucketBandwidthInetAddr"
C:\MIBS\PROJEC~1.MIB(4944) : Warning 2 : OID value for item "astaireGroup" is the same as for item "callDiversionASGroup"
C:\MIBS\PROJEC~1.MIB(5076) : Error 19 : Missing ", or }"
C:\MIBS\PROJEC~1.MIB(5148) : Error 19 : Missing ", or }"
C:\MIBS\PROJEC~1.MIB(3807) : Warning 13 : Table "mementoCallsRecordedTable" and Row "mementoCallsRecordedTableEntry" should have same name prefix
C:\MIBS\PROJEC~1.MIB(3941) : Warning 10 : Sequence "MementoSIPCassandraWriteLatencyTableEntry" and Row "mementoSIPCassandraWriteLatencyEntry" should have related names
C:\MIBS\PROJEC~1.MIB(4142) : Warning 13 : Table "mementoHTTPRequestTable" and Row "mementoHTTPRequestTableEntry" should have same name prefix
C:\MIBS\PROJEC~1.MIB(4196) : Warning 10 : Sequence "MementoHTTPRequestLatencyTableEntry" and Row "mementoHTTPRequestLatencyEntry" should have related names
C:\MIBS\PROJEC~1.MIB(4267) : Warning 10 : Sequence "MementoHTTPCassandraReadLatencyTableEntry" and Row "mementoHTTPCassandraReadLatencyEntry" should have related names
C:\MIBS\PROJEC~1.MIB(4338) : Warning 10 : Sequence "MementoRetrievedCallRecordSizeTableEntry" and Row "mementoRetrievedCallRecordSizeEntry" should have related names
[...]

PR #93, opened 2015-10-07
[Reviewer: Andy] Fix up smilint warnings

rkd-msw:
Can you review this fix, to make `smilint ./PROJECT-CLEARWATER-MIB -l 5` run cleanly?

It:
* fixes up some rogue commas
* doesn't list inaccessible objects in astaireGroup
* reworks the AstaireConnectionBucketBandwidthTable, following the guidance in http://stackoverflow.com/questions/2510211/snmp-asn-1-mib-definitions-referencing-a-table-within-a-table that nested tables should just be indexed by the parent table's index as well as their own

I'll also change Jenkins to run that smilint command and fail the build if it fails.

### DIFF
--- a/PROJECT-CLEARWATER-MIB
+++ b/PROJECT-CLEARWATER-MIB
@@ -3804,21 +3804,21 @@ callDiversionASGroup OBJECT-GROUP
 mementoSIP OBJECT IDENTIFIER ::= { memento 1 }
 
 mementoCallsRecordedTable OBJECT-TYPE
-    SYNTAX SEQUENCE OF MementoCallsRecordedTableEntry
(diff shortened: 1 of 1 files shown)


rkd-msw:
Plus a few more fixes - changing TableEntry to Entry, fixing the other warnings in #92.

PROJECT-CLEARWATER-MIB
```diff
@@ -3804,21 +3804,21 @@ callDiversionASGroup OBJECT-GROUP
 mementoSIP OBJECT IDENTIFIER ::= { memento 1 }
```
bossmc: You should change the date at the top of the file (since this is a backwards incompatible change).

bossmc:
Looks good, one comment.

PR #95, opened 2015-10-14
[Reviewer: Rob] Remove chronos from snmp handlers

eleanor-merry:


rkd-msw:
LGTM. Remove the reference from http://clearwater.readthedocs.org/en/latest/Clearwater_SNMP_Statistics/index.html too?

rkd-msw:
Actually, while you're in the area, can you remove the Homestead one as well? https://github.com/Metaswitch/homestead/pull/246/files obsoletes it.

Issue #98, opened 2015-10-27
Alarms not reliably enabled on new installs


eleanor-merry:
snmp is controlled by monit. We repeatedly call 'service snmpd restart' in clearwater-infrastructure/clearwater-snmp-handlers. As part of restart we delete the /var/run/clearwater/alarms folder, which could explain the symptoms in https://github.com/Metaswitch/clearwater-snmp-handlers/issues/96. Fix is to change the scattered snmpd restarts to snmpd stop

The alarm agent is started on install (rather than letting monit start it). This means it can get started multiple times. Fix is to set no-start in the debian/rules, and stop the alarm agent in the postinst script. 

The alarm agent is also started with an snmp_ip value of 0.0.0.0 (the default value). It's unclear so far why this is

eleanor-merry:
@JimBunch Marking as critical as this can mean that a node doesn't raise/clear any alarms

PR #99, opened 2015-10-27
[Reviewer: Matt] Control the alarm agent and snmpd through monit

eleanor-merry:
Matt, can you review these changes to keep the alarm agent and snmp under monit's control. 
Tested by installing and checking that nothing obviously broke

Part of the fix for #98

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -26,4 +26,8 @@ override_dh_auto_test:
 override_dh_auto_install:
 	mkdir debian/tmp
 
+# Don't start - monit does that for us.
+override_dh_installinit:
+	dh_installinit --no-start -u"defaults 60 40"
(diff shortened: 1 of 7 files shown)


debian/rules
```diff
@@ -26,4 +26,8 @@ override_dh_auto_test:
 override_dh_auto_install:
 	mkdir debian/tmp
 
+# Don't start - monit does that for us.
+override_dh_installinit:
+	dh_installinit --no-start -u"defaults 60 40"
```
mirw: Might be tempted to specify the package, otherwise this might apply to new packages accidentally.

mirw:
Looks good.  Presumably you've tested live?

Issue #91, opened 2015-10-02
The alarm agent should always get alarms, but not send them if snmp_ip is unset


rkd-msw:
We've tackled some bugs recently which were in alarms code, which were masked because we tested without snmp_ip set, and so all our alarms code was never run.

We would detect bugs more effectively if sprout, homestead, etc. always sent ZMQ alarms notifications to the alarm agent, and it simply ignored them if snmp_ip wasn't set.

As a bonus, this would mean the alarm agent could log out alarms even when snmp_ip wasn't set, which would make testing easier.

Discussed with @AlexHockey.

rkd-msw:
All done.

PR #100, opened 2015-10-28
[Reviewer: Alex] Use common logging and SNMP infrastructure, and log on each alarm

rkd-msw:
Part of the fix to https://github.com/Metaswitch/clearwater-snmp-handlers/issues/91.

This:

- logs to a file (/var/log/clearwater-alarms/clearwater-alarms_20151028T180000Z.txt) instead of syslog
- uses our (newish) standard SNMP setup code from cpp-common
- adds a log on each trap sent, for historical diagnostic purposes

I've tested that I get sensible logs out:

```
28-10-2015 18:39:43.361 UTC Warning alarm_trap_sender.cpp:222: Trap with alarm ID 1500.6 being sent
28-10-2015 18:39:55.017 UTC Warning alarm_trap_sender.cpp:222: Trap with alarm ID 1500.1 being sent
28-10-2015 18:40:53.825 UTC Warning alarm_trap_sender.cpp:222: Trap with alarm ID 1500.6 being sent
```

I will test that:
- alarms go out on the wire when snmp_ip is set
- alarms don't go out on the wire, and no errors are logged, when snmp_ip is unset or 0.0.0.0 - but the trap ID is still logged
- shutdown works OK (e.g. snmp_terminate doesn't hang)

### DIFF
--- a/alarms_agent.cpp
+++ b/alarms_agent.cpp
@@ -39,6 +39,7 @@
 #include <signal.h>
 #include <string>
 #include <vector>
+#include <semaphore.h>
 
 #include "utils.h"
 #include "snmp_agent.h"
--- a/alarms_agent.cpp
+++ b/alarms_agent.cpp
@@ -41,6 +41,9 @@
 #include <vector>
 
 #include "utils.h"
+#include "snmp_agent.h"
+#include "logger.h"
+#include "log.h"
 #include "alarm_table_defs.hpp"
 #include "alarm_req_listener.hpp"
 #include "alarm_model_table.hpp"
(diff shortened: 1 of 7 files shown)


alarms_agent.cpp
```diff
@@ -51,38 +54,76 @@ bool done = false;
...
-  snmp_enable_calllog();
-  snmp_enable_syslog_ident("alarms_agent", LOG_DAEMON);
-
-  // Set ourselves up as a subagent
-  netsnmp_ds_set_boolean(NETSNMP_DS_APPLICATION_ID, NETSNMP_DS_AGENT_ROLE, 1);
-  init_agent("clearwater-alarms");
+  
+  Log::setLoggingLevel(loglevel);
+  Log::setLogger(new Logger(logdir, "clearwater-alarms"));
+  TRC_ERROR("I have started");
```
AlexHockey: Seems like an odd log.

debian/clearwater-snmp-alarm-agent.init.d
```diff
@@ -87,6 +89,8 @@ get_settings()
 #
 do_start()
 {
+        mkdir -p $log_directory
+        export MIBS=""
```
AlexHockey: I have no idea what this is for. Worth a comment at least.

alarms_agent.cpp
graemerobertson: I agree that it's odd that this is an error, but I'm a big fan of writing logs in the first person. This could help us engage better with debugging problems.

alarms_agent.cpp
```diff
@@ -51,38 +54,76 @@ bool done = false;
 // Signal handler that triggers termination.
 void terminate_handler(int sig)
 {
+  snmp_terminate("clearwater-alarms");
```
AlexHockey: Is this safe to call in a signal handler?

alarms_agent.cpp
```diff
@@ -109,16 +150,14 @@ int main (int argc, char **argv)
   init_ituAlarmTable();
 
   // Run forever
-  init_snmp("clearwater-alarms");
+  init_snmp_handler_threads("clearwater-alarms");
```
AlexHockey: We used to call `netsnmp_ds_set_boolean`, `init_agent` and `init_snmp`. Now we call `snmp_setup` and `init_snmp_handler_threads`. I know next to nothing about netsnmp but that seems like quite a significant change. Is it right?

AlexHockey: Ah, this is the new SNMP setup code you alluded to. I was confused because it's been put in the global namespace and is inconsistently named (`snmp_setup` sounds like a built in netsnmp function to me). I'll stop whinging now.

AlexHockey:
Looks OK, but a few comments inline.

PR #102, opened 2015-11-10
[Reviewer: Andy] Move to being monitored by upstart, and ensure the alarm agent is sta…

### LINKED ISSUE #96: Some alarms are thrown way on boot because snmpd/alarm-agent isn't up yet
On boot I see messages like this in syslog:
```
dropped request: 'issue-alarm cluster-manager 8001.4'
```

I also see the alarm-agent getting a zmq_bind error (```No such file or directory```) on ```ipc:///var/run/clearwater/alarms```. My workaround code of having the alarm-agent kill itself prevents the alarm-agent from being stuck in that state, but it's worth someone else taking a look at why it (sometimes) gets that error when 0MQ is supposed to create the socket (and does at other times). I didn't have the time to get to the bottom of it, hence the workaround.

Diags are available

rkd-msw:
…rted before monit

Should fix #96.


bossmc:
Looks good.

PR #103, opened 2015-11-10
[Reviewer: Graeme] Don't install nonexistent monit file

rkd-msw:


graemerobertson:
Could be right, no idea what your previous change was. +1

PR #105, opened 2015-11-18
[Reviewer: Andy] MIB with u32 changes in

johadalin:
MIBMIBMIB

Really nothing controversial. Unsigned32 seems to work on brief live tests. will test further tomorrow. 
Passes smilint without error.


bossmc:
Sure.

PR #111, opened 2015-12-14
[Reviewer: Ellie] Fixes for issue #96

rkd-msw:
- create the /var/run/clearwater directory
- catch zmq_bind failures before starting the listener thread
- fire the termination semaphore when the listener thread exits

This doesn't fully implement Monit polling of the alarm agent, but does add a 'poll' message to the list of ZMQ messages we'll accept.

### DIFF
--- a/alarm_req_listener.cpp
+++ b/alarm_req_listener.cpp
@@ -50,9 +50,15 @@
 
 AlarmReqListener AlarmReqListener::_instance;
 
-bool AlarmReqListener::start()
+bool AlarmReqListener::start(sem_t* term_sem)
 {
+  _term_sem = term_sem;
   if (!zmq_init_ctx())
+  {
+    return false;
+  }
+
+  if (!zmq_init_sck())
   {
     return false;
   }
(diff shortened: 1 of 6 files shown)


alarm_req_listener.cpp
```diff
@@ -238,6 +242,10 @@ void AlarmReqListener::listener()
     {
       AlarmTrapSender::get_instance().sync_alarms(false);
     }
+    else if ((msg[0].compare("poll") == 0) && (msg.size() == 1))
+    {
+      // do nothing, just reply "ok"
```
eleanor-merry: Do you need to do anything here to reply OK?

rkd-msw: Line 254 does that.

alarms_agent.cpp
```diff
@@ -95,7 +95,6 @@ int main (int argc, char **argv)
   int c;
   int optind;
 
-  printf("Helo world\n");
```
eleanor-merry: Nice!

eleanor-merry:
LGTM

PR #112, opened 2015-12-14
Implementing Active Alarm Table

AME2:


alarms_agent.cpp
```diff
@@ -49,11 +49,12 @@
 #include "alarm_req_listener.hpp"
 #include "alarm_model_table.hpp"
 #include "itu_alarm_table.hpp"
+#include "alarm_active_table.hpp"
 
 static sem_t term_sem;
-
+std::string local_ip = "0.0.0.0";
```
rkd-msw: Should this be inside `main`, like other command-line arguments?

alarm_trap_sender.hpp
```diff
@@ -170,7 +184,7 @@ class AlarmTrapSender
   void sync_alarms(bool do_clear);
 
   static AlarmTrapSender& get_instance() {return _instance;}
-
+  alarmActiveTable_SNMPDateTime* alarm_time_issued(void);
```
rkd-msw: As it's only the active alarm table that needs the time in this precise format, should this be part of the active alarm table rather than part of the trap sender?

alarm_trap_sender.cpp
```diff
@@ -157,7 +193,14 @@ void AlarmTrapSender::issue_alarm(const std::string& issuer, const std::string&
     {
       if (!AlarmFilter::get_instance().alarm_filtered(index, severity))
       {
+        alarmActiveTable_create_row((char*) "", AlarmTrapSender::alarm_time_issued(), alarm_counter, alarm_table_def);
         send_trap(alarm_table_def);
+        alarm_counter++;
+        //Reset the alarm counter incramentor if it reaches its max limit
+        if (alarm_counter == 4294967295)
```
rkd-msw: `UINT_MAX` is often a clearer way to express 4294967295.

alarm_trap_sender.cpp
```diff
@@ -157,7 +193,14 @@ void AlarmTrapSender::issue_alarm(const std::string& issuer, const std::string&
     {
       if (!AlarmFilter::get_instance().alarm_filtered(index, severity))
       {
+        alarmActiveTable_create_row((char*) "", AlarmTrapSender::alarm_time_issued(), alarm_counter, alarm_table_def);
         send_trap(alarm_table_def);
+        alarm_counter++;
+        //Reset the alarm counter incramentor if it reaches its max limit
```
rkd-msw: You don't need the word "incrementor" in this sentence - 'Reset the alarm counter' makes sense by itself - but you do want a space after the `//`.

alarm_trap_sender.cpp
```diff
@@ -157,7 +193,14 @@ void AlarmTrapSender::issue_alarm(const std::string& issuer, const std::string&
     {
       if (!AlarmFilter::get_instance().alarm_filtered(index, severity))
       {
+        alarmActiveTable_create_row((char*) "", AlarmTrapSender::alarm_time_issued(), alarm_counter, alarm_table_def);
```
rkd-msw: If you declare the first parameter of `alarmActiveTable_create_row` as a `const char*`, you don't need this cast. (Alternatively, as it's always the same, you could do away with it altogether.)

alarm_trap_sender.cpp
```diff
@@ -138,8 +142,40 @@ unsigned long AlarmFilter::current_time_ms()
...
+  d_time->direction = plus_or_minus;
+  d_time->timegeozone = daylight_savings;
+  d_time->mintimezone = (abs(timezone) % 3600) / 60;
+
+  return d_time;
+}
+
 void AlarmTrapSender::issue_alarm(const std::string& issuer, const std::string& identifier)
 {
+  static unsigned int alarm_counter = 1;
```
rkd-msw: Rather than static local variables in functions, it's cleaner to use a member variable in the AlarmTrapSender class.

alarm_trap_sender.cpp
```diff
@@ -138,8 +142,40 @@ unsigned long AlarmFilter::current_time_ms()
...
+  int daylight_savings;
+
+  clock_gettime(CLOCK_REALTIME, &ts);
+  rawtime = ts.tv_sec;
+  timing = gmtime(&rawtime);
+  // This updates the variable timezone declared within time.h which
+  // tells us how offset (in seconds) from UTC we are.
+  localtime(&rawtime);
+  plus_or_minus = (timezone > 0) ? '+' : '-';
+  daylight_savings = abs(timezone) / 3600;
```
rkd-msw: This isn't the right name for this variable - it's the offset from UTC, not the flag indicating whether daylight savings is in effect. (For example, I think this will always be 6 in Bangladesh, which doesn't have daylight savings time.)

alarm_trap_sender.cpp
```diff
@@ -138,8 +142,40 @@ unsigned long AlarmFilter::current_time_ms()
...
+  d_time->year = htons(1900 + timing->tm_year);
```
rkd-msw: Worth noting why we're using htons here (referencing the RFC that talks about it being in network byte order).

alarm_trap_sender.cpp
```diff
@@ -138,8 +142,40 @@ unsigned long AlarmFilter::current_time_ms()
...
+  d_time->month = 1 + timing->tm_mon;
+  d_time->day = timing->tm_mday;
+  d_time->hour = timing->tm_hour;
+  d_time->minute = timing->tm_min;
+  d_time->second = timing->tm_sec;
+  d_time->decisecond = 0;
+  d_time->direction = plus_or_minus;
```
rkd-msw: I'm not sure a separate plus_or_minus variable makes things clearer here - I'd just write `d_time->direction = (timezone > 0) ? '+' : '-';`.

alarm_trap_sender.cpp
```diff
@@ -138,8 +142,40 @@ unsigned long AlarmFilter::current_time_ms()
...
+  time_t rawtime;
+  char plus_or_minus;
+  int daylight_savings;
+
+  clock_gettime(CLOCK_REALTIME, &ts);
+  rawtime = ts.tv_sec;
+  timing = gmtime(&rawtime);
+  // This updates the variable timezone declared within time.h which
+  // tells us how offset (in seconds) from UTC we are.
+  localtime(&rawtime);
```
rkd-msw: I think you should be setting `timing` from `localtime`, not from `gmtime` - otherwise you're using the UTC hours and seconds, even when you're specifying an offset from UTC.

alarm_active_table.hpp
```diff
@@ -0,0 +1,128 @@
...
+ *
+ *  oid declarations
+ */
+extern oid alarmActiveTable_oid[];
+extern size_t alarmActiveTable_oid_len;
+
+#define alarmModelTable_TABLE_OID 1,3,6,1,2,1,118,1,1,2
+#define alarmActiveTable_TABLE_OID 1,3,6,1,2,1,118,1,2,2
+    
+#define ZERO_DOT_ZERO_OID 0,0
```
rkd-msw: Should these be moved out to a separate, common header file, rather than being duplicated like this?

alarm_active_table.hpp
```diff
@@ -0,0 +1,128 @@
...
+#define ALARM_ACTIVE_STATE_OID 1,3,6,1,2,1,118,0,2
+#define ALARM_CLEAR_STATE_OID 1,3,6,1,2,1,118,0,3
+
+#define ENTRY_COLUMN_OID 1,3
+
+/*************************************************************
+ *
+ *  column number definitions for table alarmModelTable
```
rkd-msw: "...for table alarmActiveTable"?

alarm_active_table.hpp
```diff
@@ -0,0 +1,128 @@
...
+typedef struct alarmActiveTable_context_s {
+    netsnmp_index _index; /** THIS MUST BE FIRST!!! */
+
+    AlarmTableDef* _alarm_table_def;
+
+} alarmActiveTable_context;
+
+/*************************************************************
+ *
+ *  function declarations
```
rkd-msw: I'm not sure this comment is necessary - it's pretty obvious that these are function declarations.

alarm_active_table.cpp
```diff
@@ -0,0 +1,423 @@
...
+ *
+ *  $Id:$
+ */
+
+#include "alarm_active_table.hpp"
+#include "log.h"
+
+#include <string>
+
+using namespace std;
```
rkd-msw: We don't normally use `using namespace std;` (and I think your code works fine without it).

alarm_active_table.cpp
```diff
@@ -0,0 +1,423 @@
...
+
+static netsnmp_handler_registration* my_handler = NULL;
+static netsnmp_table_array_callbacks cb;
+
+static oid alarmModelTable_oid[] = { alarmModelTable_TABLE_OID };
```
rkd-msw: Worth moving these out to a common file, rather than duplicating them?

alarm_active_table.cpp
```diff
@@ -0,0 +1,423 @@
...
+size_t alarmActiveTable_oid_len = OID_LENGTH(alarmActiveTable_oid);
+
+static oid zero_dot_zero_oid[] = { ZERO_DOT_ZERO_OID };
+static oid alarm_active_state_oid[] = { ALARM_ACTIVE_STATE_OID };
+static oid alarm_clear_state_oid[] = { ALARM_CLEAR_STATE_OID };
+
+static oid entry_column_oid[] = { ENTRY_COLUMN_OID };
+static size_t entry_column_oid_len = OID_LENGTH(entry_column_oid);
+
+std::map<unsigned long, alarmActiveTable_context*> pointers;
```
rkd-msw: `pointers` isn't a great name for this - something like `alarm_indexes_to_rows` would be clearer, especially if you also had a comment explaining what this was for.

alarm_active_table.cpp
```diff
@@ -0,0 +1,423 @@
...
+std::string loc_ip;
```
rkd-msw: `local_ip`?

alarm_active_table.cpp
```diff
@@ -0,0 +1,423 @@
...
+
+/************************************************************
+ *
+ *  Initialize the alarmActiveTable table by defining its contents
```
rkd-msw: Unlike alarmModelTable, I don't think this function does define the table's contents, so this comment seems inaccurate.

alarm_active_table.cpp
```diff
@@ -0,0 +1,423 @@
...
+  return SNMP_ERR_NOERROR;
+}
+
+/************************************************************
+ *
+ *  This routine is called for get requests to copy the data
+ *  from the context to the varbind for the request.
+ */
+int alarmActiveTable_get_value(netsnmp_request_info* request,
+                              netsnmp_index* item,
```
rkd-msw: The indentation here is one character off.

alarm_active_table.cpp
```diff
@@ -0,0 +1,423 @@
...
+                              netsnmp_table_request_info* table_info)
+{
+  netsnmp_variable_list* var = request->requestvb;
+  alarmActiveTable_context* context = (alarmActiveTable_context*) item;
+  switch(table_info->colnum)
```
rkd-msw: Normally we'd have a space here - `switch (table_info->colnum)` - like for if statements.

alarm_active_table.cpp
```diff
@@ -0,0 +1,423 @@
...
+  {
+    case COLUMN_ALARMACTIVEENGINEID:
+    {
+      // Cannot obtain with SNMPv2 so use zero-length-string //
+      static std::string empty_string  = "";
```
rkd-msw: As you have `static std::string empty_string  = "";` in a couple of places, should you define it at the top of the file, as you do with the 0.0 OID?

alarm_active_table.cpp
```diff
@@ -0,0 +1,423 @@
...
+      snmp_set_var_typed_value(var, ASN_OCTET_STR,
+                               (u_char*) &empty_string,
```
rkd-msw: Should technically be `empty_string.c_str()`, although as you're reading no bytes from it, that's pretty minor.

alarm_active_table.cpp
```diff
@@ -0,0 +1,423 @@
...
+                               empty_string.length());
+    }
+    break;
+
+    case COLUMN_ALARMACTIVEENGINEADDRESSTYPE:
+    {
+      // This is of type InetAddressType and will either be
+      // IPv4 or IPv6
+      int ip_type;
```
rkd-msw: You don't need to declare and assign these separately - `int ip_type = (loc_ip.find(":") == std::string::npos) ? 1 : 2;` is fine.

alarm_active_table.cpp
```diff
@@ -0,0 +1,423 @@
...
+      ip_type = (loc_ip.find(":") == std::string::npos) ? 1 : 2;
```
rkd-msw: It's not immediately obvious what 1 and 2 are here - a comment referring to the RFC, and/or defining them as `const static int SNMP_IPv4_ADDRTYPE = 1`, would help.

alarm_active_table.cpp
```diff
@@ -0,0 +1,423 @@
...
+                               (u_char*) &empty_string,
+                               empty_string.length());
+    }
+    break;
+    
+    case COLUMN_ALARMACTIVEVARIABLES:
+    {
+      // The number of variables in alarmActiveVariableTable for
+      // this alarm. See as we have no such table this will be 0.
+      static unsigned long variables = 0;
```
rkd-msw: I don't think this needs to be static. (It also doesn't need to be a long - longs will be 8 bytes, but an ASN_UNSIGNED is 4, so I'd use a uint32_t to match.)

alarm_active_table.cpp
```diff
@@ -0,0 +1,423 @@
...
+                               (u_char*) &variables,
+                               sizeof(variables));
+    }
+    break;
+
+    case COLUMN_ALARMACTIVENOTIFICATIONID:
+    {
+      if (context->_alarm_table_def->severity() == AlarmDef::CLEARED)
+      {
+        snmp_set_var_typed_value(var, ASN_OBJECT_ID,
```
rkd-msw: I think we'd agreed that this will always be alarm_active_state_oid, so you can simplify this a bit.

alarm_active_table.cpp
```diff
@@ -0,0 +1,423 @@
...
+      snmp_set_var_typed_value(var, ASN_OBJECT_ID,
+                               (u_char*) zero_dot_zero_oid,
+                               sizeof(zero_dot_zero_oid));
+    }
+    break;
+
+    case COLUMN_ALARMACTIVEMODELPOINTER:
+    {
+      oid* model_pointer;
+      // Convert the index string OID
```
rkd-msw: The next line isn't converting the index string oid.

alarm_active_table.cpp
```diff
@@ -0,0 +1,423 @@
...
+      int model_pointer_len = alarmModelTable_oid_len + entry_column_oid_len + 3;
```
rkd-msw: Worth noting why you have the 3 (to accommodate the 3 index OIDs of the model table).

alarm_active_table.cpp
```diff
@@ -0,0 +1,423 @@
...
+      // Append the index array on to the Alarm Model Table OID, inserting
+      // the OID ".1.3." for the entry in the first non index column
+      model_pointer = new oid[model_pointer_len];
+      copy(alarmModelTable_oid, alarmModelTable_oid + alarmModelTable_oid_len, model_pointer);
+
+      copy(entry_column_oid, entry_column_oid + entry_column_oid_len, model_pointer + alarmModelTable_oid_len);
+      
+      model_pointer[alarmModelTable_oid_len + entry_column_oid_len] = 0;
```
rkd-msw: Would it be simpler just to add the `.1.3` for the entry columns like this? (I don't feel strongly.)

alarm_active_table.cpp
```diff
@@ -0,0 +1,423 @@
...
+    break;
+
+    case COLUMN_ALARMACTIVEMODELPOINTER:
+    {
+      oid* model_pointer;
+      // Convert the index string OID
+      int model_pointer_len = alarmModelTable_oid_len + entry_column_oid_len + 3;
+      // Append the index array on to the Alarm Model Table OID, inserting
+      // the OID ".1.3." for the entry in the first non index column
+      model_pointer = new oid[model_pointer_len];
```
rkd-msw: I think this probably leaks memory - you're calling `new` without a corresponding `delete`.

alarm_active_table.cpp
```diff
@@ -0,0 +1,423 @@
...
+  return SNMP_ERR_NOERROR;
+}
+
+/************************************************************
+ * 
+ *  Create a new row context and initialize its oid index.
+ */
+void alarmActiveTable_create_row(char* name,
+                                 alarmActiveTable_SNMPDateTime* datetime,
+                                 unsigned long index,
```
rkd-msw: This is probably better named as `alarm_counter`. (Should the alarm table be responsible for the alarm counter, rather than the trap sender being responsible for it and passing it in?).

alarm_active_table.cpp
```diff
@@ -0,0 +1,423 @@
...
+  alarmActiveTable_context* ctx = SNMP_MALLOC_TYPEDEF(alarmActiveTable_context);
+  ctx->_alarm_table_def = &def;
+  if (pointers[ctx->_alarm_table_def->alarm_index()])
+  {
+    if (ctx->_alarm_table_def->severity() == pointers[ctx->_alarm_table_def->alarm_index()]->_alarm_table_def->severity())
+    {
+      return;
+    }
+  }
+  if (!ctx)
```
rkd-msw: I think if you hit this error condition, you'll have segfaulted on line 298 above, when you assign to `ctx->_alarm_table_def` even though ctx is NULL.

alarm_active_table.cpp
```diff
@@ -0,0 +1,423 @@
...
+ *  Create a new row context and initialize its oid index.
+ */
+void alarmActiveTable_create_row(char* name,
+                                 alarmActiveTable_SNMPDateTime* datetime,
+                                 unsigned long index,
+                                 AlarmTableDef& def)
+{
+  alarmActiveTable_context* ctx = SNMP_MALLOC_TYPEDEF(alarmActiveTable_context);
+  ctx->_alarm_table_def = &def;
+  if (pointers[ctx->_alarm_table_def->alarm_index()])
```
rkd-msw: It's not very clear what this code is doing. Comments would definitely help, as would clearer variable names:

```
alarmActiveTable_context* existing_row = alarm_indexes_to_rows[def.alarm_index()];
if (existing_row != NULL)
{
  if (def.severity == existing_row->_alarm_table_def->severity())
  {
    // Existing alarm has been re-raised, do nothing
    return;
  }
}
```

seems a bit clearer to me.

alarm_active_table.cpp
```diff
@@ -0,0 +1,423 @@
...
+  }
+  if (ctx->_alarm_table_def->severity() == 1)
+  {
+    alarmActiveTable_delete_row(def);
+  }
+  else
+  {
+    if (pointers[ctx->_alarm_table_def->alarm_index()])
+    {
+      CONTAINER_REMOVE(cb.container, pointers[ctx->_alarm_table_def->alarm_index()]);
```
rkd-msw: This feels like it makes more sense above (alongside your other check for the existing row).

alarm_active_table.cpp
```diff
@@ -0,0 +1,423 @@
...
+    }
+    
+    if (ctx)
+    {
+      CONTAINER_INSERT(cb.container, ctx);
+      pointers[ctx->_alarm_table_def->alarm_index()] = ctx;
```
rkd-msw: Does this leak memory (specifically, the previously-allocated alarmActiveTable_context which might have been here)?

alarm_active_table.cpp
```diff
@@ -0,0 +1,423 @@
...
+  if (!ctx)
+  {
+    TRC_ERROR("malloc failed: alarmActiveTable_remove_row");
+    return;
+  }
+  
+  if (ctx)
+  {
+    CONTAINER_REMOVE(cb.container, pointers[ctx->_alarm_table_def->alarm_index()]);
+    pointers[ctx->_alarm_table_def->alarm_index()] = NULL;
```
rkd-msw: As above, I think this leaks memory and you should be calling `free()`.

alarm_active_table.cpp
```diff
@@ -0,0 +1,423 @@
...
+  return;
+}
+
+/************************************************************
+ *
+ * Delete a row
+ */
+void alarmActiveTable_delete_row(AlarmTableDef& def)
+{
+  alarmActiveTable_context* ctx = SNMP_MALLOC_TYPEDEF(alarmActiveTable_context);
```
rkd-msw: Why are you allocating a new alarmActiveTable_context at this point?

alarm_active_table.cpp
```diff
@@ -0,0 +1,423 @@
...
+ *
+ *  Convert table index components to an oid.
+ */
+int alarmActiveTable_index_to_oid(char* name,
+                                 alarmActiveTable_SNMPDateTime* datetime,
+                                 unsigned long counter,
+                                 netsnmp_index *oid_idx)
+{
+  int err = SNMP_ERR_NOERROR;
+
```
rkd-msw: Extra whitespace.

alarm_active_table.cpp
```diff
@@ -0,0 +1,423 @@
...
+
+  netsnmp_variable_list var_alarmListName;
+  netsnmp_variable_list var_alarmActiveDateAndTime;
+  netsnmp_variable_list var_alarmActiveIndex;
```
rkd-msw: You should be able to make this area more succinct by using [designated initializers](https://gcc.gnu.org/onlinedocs/gcc/Designated-Inits.html):

`netsnmp_variable_list var_alarmActiveIndex = {.type = ASN_UNSIGNED, .next_variable = NULL};`

alarm_active_table.cpp
```diff
@@ -0,0 +1,423 @@
...
+
+
+  DEBUGMSGTL(("verbose:alarmActiveTable:alarmActiveTable_index_to_oid", "called\n"));
+
+  snmp_set_var_value(&var_alarmListName, (u_char*) name, strlen(name));
+  //Here we use the value 11 as that is how many octets the datetime structure
+  //contains. We were not able to find the size of the structure using sizeof
+  //due to structure padding.
+  snmp_set_var_value(&var_alarmActiveDateAndTime, (u_char*) datetime, 11);
+  TRC_STATUS("Counter: %p, %d %d %d %d size %d", &counter,
```
rkd-msw: I think this is probably left over from debugging?

alarm_active_table.cpp
```diff
@@ -0,0 +1,423 @@
...
+
+  /*
+   * chain index varbinds together
+   */
+  var_alarmListName.next_variable = &var_alarmActiveDateAndTime;
+  var_alarmActiveDateAndTime.next_variable = &var_alarmActiveIndex;
+  var_alarmActiveIndex.next_variable =  NULL;
+
+
+  DEBUGMSGTL(("verbose:alarmActiveTable:alarmActiveTable_index_to_oid", "called\n"));
```
rkd-msw: Is this a useful log? If so, should it be TRC_DEBUG?

alarm_active_table.cpp
```diff
@@ -0,0 +1,423 @@
...
+  {
+    if (ctx->_index.oids != NULL)
+    {
+      free(ctx->_index.oids);
+    }
+
+    free(ctx);
+    return;
+  }
+  if (ctx->_alarm_table_def->severity() == 1)
```
rkd-msw: Should this be `== AlarmDef::CLEARED` as above?

alarm_active_table.cpp
```diff
@@ -0,0 +1,423 @@
...
+      return SNMP_ERR_GENERR;
+    }
+  }
+
+  return SNMP_ERR_NOERROR;
+}
+
+/************************************************************
+ * 
+ *  Create a new row context and initialize its oid index.
```
rkd-msw: I think this function actually does more than this comment says (e.g. it's responsible for ensuring there's at most one row for each alarm).

rkd-msw:
Generally looks good - I had some comments about coding style, clarity, and possible memory leaks.

A couple of things that didn't make sense as inline comments:

- the pull request title ("Alarms") is a bit generic - I'm not sure I'd be easily able to find this in future if I were looking through https://github.com/Metaswitch/clearwater-snmp-handlers/pulls?q=is%3Apr+is%3Aclosed
- you've added an empty alarm_agent.cpp and alarms_agent.hpp - can you `git rm` these?

and a couple of high-level themes from my feedback:

- you use static variables in functions more often than we normally do - generally, if data needs to last beyond the lifetime of a function, it should be a member variable (if it varies) or just an ordinary constant
- there were a couple of possible memory leaks - generally, every time you allocate some memory with malloc or new, you should know where it should be freed again and make sure that happens (especially in code like this, where you don't have UTs to catch it easily)
- there were some places where I wasn't sure code was in the right place - if you're only keeping a counter, or representing dates in a particular way, because of the requirements of the active alarm table, it makes sense to keep that code inside the active alarm table
- it's important to make sure that your code is as clear as possible for programmers in future - where three ways to do this are to avoid magic numbers (e.g. prefer `AlarmDef::CLEARED ` instead of `1`), to choose good variable names, and to make sure particularly tricky bits of the code (e.g. https://github.com/Metaswitch/clearwater-snmp-handlers/pull/112/files#r47566795) are well-commented.

alarm_trap_sender.hpp
```diff
@@ -39,9 +39,23 @@
 
 #include <string>
 #include <map>
+#include <cstdint>
 
 #include "alarm_table_defs.hpp"
 
+typedef struct SNMPDateTime {
```
rkd-msw: It would be good to have a comment referencing the RFC - particularly to make it clear to anyone modifying this code in future that this isn't an arbitrary format, it's carefully written to match precise byte layout requirements.

alarm_active_table.cpp
AME2: I don't think this is useful. I think this is just telling us the function alarmActiveTable_index_to_oid has been called so I suggest removing this line altogether.

alarm_active_table.cpp
```diff
@@ -0,0 +1,453 @@
...
+  /*
+   * registering the table with the master agent
+   */
+  cb.get_value = alarmActiveTable_get_value;
+  cb.container = netsnmp_container_find("alarmActiveTable_primary:"
+                                        "alarmActiveTable:"
+                                        "table_container");
+  cb.can_set = 0;
+
+  DEBUGMSGTL(("initialize_table_alarmActiveTable", "Registering as table array\n"));
```
rkd-msw: Like the other DEBUGMSGTL I commented on, can you change this to TRC_DEBUG?

alarm_active_table.cpp
```diff
@@ -0,0 +1,453 @@
...
+/*************************************************************
+ *
+ * Find the current time and use it to fill the d_time struct
+ * which is formatted to be compliant with RFC 3877
+ */
+alarmActiveTable_SNMPDateTime alarm_time_issued(void)
+{
+  alarmActiveTable_SNMPDateTime d_time;
+  struct timespec ts;
+  struct tm *timing = new tm;
```
rkd-msw: I think you could simplify this by just declaring `struct tm timing;` on the stack, then using `&timing` when you need a pointer to pass into localtime_r.

alarm_active_table.cpp
```diff
@@ -0,0 +1,453 @@
...
+  time_t rawtime;
+  int UTC_offset;
+
+  clock_gettime(CLOCK_REALTIME, &ts);
+  rawtime = ts.tv_sec;
+  // localtime_r converts calendar time to a broken down representation
+  // to be stored in the stucture timing
+  localtime_r(&rawtime, timing);
```
rkd-msw: I don't think you need to call both `localtime` and `localtime_r` - that looks odd. (I see that `localtime_r` "need not set tzname, timezone, and daylight", which you need, so I'd probably just call localtime, with a comment explaining why we can't call `localtime_r`).

alarm_active_table.cpp
```diff
@@ -0,0 +1,453 @@
...
+ * is raised. This ensures each alarm has at most one entry
+ * in Alarm Active Table.
+ */
+void alarmActiveTable_trap_handler(AlarmTableDef& def)
+{
+  alarmActiveTable_context* existing_row = alarm_indexes_to_rows[def.alarm_index()];
+  // 1) New alarm
+  if (existing_row == NULL && def.severity() != AlarmDef::CLEARED)
+  {
+  alarmActiveTable_create_row(def);
```
rkd-msw: This line (and all the bodies of if statements) should be indented.

alarm_active_table.cpp
```diff
@@ -0,0 +1,453 @@
...
+  // 2) Existing alarm with different severity
+  else if (existing_row != NULL && def.severity() != AlarmDef::CLEARED && def.severity() != existing_row->_alarm_table_def->severity())
+  {
+  alarmActiveTable_delete_row(def);
+  alarmActiveTable_create_row(def);
+  }
+  // 3) Existing alarm with same severity
+  else if (existing_row != NULL && def.severity() != AlarmDef::CLEARED && def.severity() == existing_row->_alarm_table_def->severity())
+  {
+  return;
```
rkd-msw: I probably wouldn't return at this point - as a general rule, it's good to have as few return statements in a function as possible (so that if you want to free memory, or log a message, before returning, you have fewer places to do it and less chance of missing one). I'd either have a comment just saying `// No-op`, or exclude this case from the if statement entirely (so it only had the three cases where we need to do something - raising, updating or clearing an alarm).

alarm_active_table.cpp
```diff
@@ -0,0 +1,453 @@
...
+  var_alarmActiveIndex.type = ASN_UNSIGNED;
+
+  /*
+   * chain index varbinds together
+   */
+  var_alarmListName.next_variable = &var_alarmActiveDateAndTime;
+  var_alarmActiveDateAndTime.next_variable = &var_alarmActiveIndex;
+  var_alarmActiveIndex.next_variable =  NULL;
+  
+  DEBUGMSGTL(("verbose:alarmActiveTable:alarmActiveTable_index_to_oid", "called\n"));
```
rkd-msw: TRC_DEBUG (or remove this entirely?).

alarm_active_table.hpp
```diff
@@ -0,0 +1,115 @@
...
+
+#include <alarm_table_defs.hpp>
+#include "alarm_trap_sender.hpp"
+#include "oid_definitions.hpp"
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <string>
+using namespace std;
```
rkd-msw: We don't normally use `using namespace std;` - you should remove it from here too.

rkd-msw:
One thing to note - it looks like you've changed the build-infra and cpp-common references. You should make sure that both of these are right (i.e. the checked-in submodule reference is to 'master', and isn't different from the reference on the master branch of clearwater-snmp-handlers).

```
cd build-infra
git fetch && git checkout origin/master
cd ../modules/cpp-common
git fetch && git checkout origin/master
cd ../..
git add build-infra modules/cpp-common
git commit
```

should do this.

Otherwise, looks good - feel free to merge once you've made those markups.

PR #113, opened 2015-12-17
Implementing Alarm Active Table: Fixing UTs

AME2:


alarm_active_table.cpp
```diff
@@ -107,6 +107,7 @@ int init_alarmActiveTable(std::string ip)
   cb.container = netsnmp_container_find("alarmActiveTable_primary:"
                                         "alarmActiveTable:"
                                         "table_container");
+  printf("RKD: Value of cb.container %p\n", cb.container);
```
rkd-msw: You should remove this printf.

alarm_active_table.cpp
```diff
@@ -364,6 +369,7 @@ void alarmActiveTable_create_row(AlarmTableDef& def)
   }
   if (ctx)
   {
+    printf("RKD: %p, %p\n", cb.container, ctx);
```
rkd-msw: You should remove this printf.

ut/alarm_req_listener_test.cpp
```diff
@@ -43,6 +43,7 @@
 #include "alarm.h"
 #include "alarmdefinition.h"
 
+#include "alarm_active_table.hpp"
```
rkd-msw: Do you need this?

alarm_active_table.cpp
```diff
@@ -307,32 +308,36 @@ alarmActiveTable_SNMPDateTime alarm_time_issued(void)
...
-    // No-op
-  }
-  // 4) Clearing a raised alarm
-  else if (existing_row != NULL && def.severity() == AlarmDef::CLEARED)
-  {
-    alarmActiveTable_delete_row(def);
-  }
-  // 5) Clearing an unraised alarm 
-  else if (existing_row == NULL && def.severity() == AlarmDef::CLEARED)
+  // This stops UTs from failing due to mocking up the cb object
```
rkd-msw: I might say "from failing when the alarm table is uninitialized" rather than "from failing due to mocking up the cb object" - I think that's a bit clearer to future readers.

rkd-msw:
Couple of comments - as discussed, it's worth looking over the pull request before you assign it to me, as you'll be just as able to spot rogue printf lines as I will.

One other thing to note is that we usually put the reviewer in the pull request title (so that they can spot it at a glance) - e.g. as I've done in https://github.com/Metaswitch/clearwater-sas-resources/pull/82.

PR #115, opened 2016-01-04
[Reviewer: Rob] Creates a custom alarm definition

AME2:
Also see changes made in cpp-common:
https://github.com/Metaswitch/cpp-common/pull/438

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,7 @@ cw_alarm_agent_SOURCES := alarms_agent.cpp snmp_agent.cpp ${AGENT_COMMON_SOURCES
 cw_alarm_test_SOURCES := test_main.cpp \
                          alarm.cpp \
                          alarm_table_defs_test.cpp \
+                         alarm_tables_testing.cpp \
                          alarm_req_listener_test.cpp \
                          test_interposer.cpp \
                          fakenetsnmp.cpp \
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,6 @@ cw_alarm_fvtest_SOURCES := test_main.cpp \
                            alarm.cpp \
                            alarm_tables_testing.cpp\
                            ${AGENT_COMMON_SOURCES}
-test: run_cw_alarm_test
 fvtest: run_cw_alarm_fvtest
 AGENT_COMMON_CPPFLAGS := -I. \
                          -Imodules/cpp-common/include \
@@ -36,8 +35,8 @@ cw_alarm_agent_CPPFLAGS := ${AGENT_COMMON_CPPFLAGS}
 cw_alarm_test_CPPFLAGS := ${AGENT_COMMON_CPPFLAGS}
 cw_alarm_fvtest_CPPFLAGS := ${AGENT_COMMON_CPPFLAGS}
 
-cw_alarm_test_COVERAGE_EXCLUSIONS := ^modules/rapidjson|^modules/cpp-common/test_utils|^modules/cpp-common/include
-cw_alarm_fvtest_COVERAGE_EXCLUSIONS := ^modules/rapidjson|^modules/cpp-common/test_utils|^modules/cpp-common/include
+cw_alarm_test_COVERAGE_EXCLUSIONS := ^modules/rapidjson|^modules/cpp-common/test_utils|^modules/cpp-common/include|^modules/cpp-common/src|alarm_active_table.cpp|alarm_model_table.cpp|itu_alarm_table.cpp
+cw_alarm_fvtest_COVERAGE_EXCLUSIONS := ^modules/rapidjson|^modules/cpp-common/test_utils|^modules/cpp-common/include|^modules/cpp-common/src|alarm_req_listener.cpp|alarm_trap_sender.cpp|alarm_trap_sender.hpp
 
 AGENT_COMMON_LDFLAGS := -lzmq \
                         -lpthread \
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 TARGETS := cw_alarm_agent cdiv_handler.so memento_as_handler.so memento_handler.so astaire_handler.so
-TEST_TARGETS := cw_alarm_test
+TEST_TARGETS := cw_alarm_test cw_alarm_fvtest
 
 CPPFLAGS_TEST += -Imodules/cpp-common/test_utils
 
@@ -17,21 +17,27 @@ cw_alarm_agent_SOURCES := alarms_agent.cpp snmp_agent.cpp ${AGENT_COMMON_SOURCES
 cw_alarm_test_SOURCES := test_main.cpp \
                          alarm.cpp \
                          alarm_table_defs_test.cpp \
-                         alarm_tables_testing.cpp \
                          alarm_req_listener_test.cpp \
                          test_interposer.cpp \
                          fakenetsnmp.cpp \
                          fakelogger.cpp \
                          fakezmq.cpp \
                          ${AGENT_COMMON_SOURCES}
-
+cw_alarm_fvtest_SOURCES := test_main.cpp \
+                           alarm.cpp \
+                           alarm_tables_testing.cpp\
+                           ${AGENT_COMMON_SOURCES}
+test: run_cw_alarm_test
+fvtest: run_cw_alarm_fvtest
 AGENT_COMMON_CPPFLAGS := -I. \
                          -Imodules/cpp-common/include \
                          -Imodules/rapidjson/include
 cw_alarm_agent_CPPFLAGS := ${AGENT_COMMON_CPPFLAGS}
 cw_alarm_test_CPPFLAGS := ${AGENT_COMMON_CPPFLAGS}
+cw_alarm_fvtest_CPPFLAGS := ${AGENT_COMMON_CPPFLAGS}
 
 cw_alarm_test_COVERAGE_EXCLUSIONS := ^modules/rapidjson|^modules/cpp-common/test_utils|^modules/cpp-common/include
+cw_alarm_fvtest_COVERAGE_EXCLUSIONS := ^modules/rapidjson|^modules/cpp-common/test_utils|^modules/cpp-common/include
 
 AGENT_COMMON_LDFLAGS := -lzmq \
                         -lpthread \
@@ -41,6 +47,7 @@ AGENT_COMMON_LDFLAGS := -lzmq \
                         `net-snmp-config --agent-libs`
 cw_alarm_agent_LDFLAGS := ${AGENT_COMMON_LDFLAGS}
 cw_alarm_test_LDFLAGS := ${AGENT_COMMON_LDFLAGS}
+cw_alarm_fvtest_LDFLAGS := ${AGENT_COMMON_LDFLAGS}
 
 PLUGINS_COMMON_SOURCES := custom_handler.cpp oid.cpp oidtree.cpp oid_inet_addr.cpp zmq_listener.cpp zmq_message_handler.cpp
 cdiv_handler.so_SOURCES := cdivdata.cpp ${PLUGINS_COMMON_SOURCES}
@@ -71,6 +78,7 @@ include build-infra/cpp.mk
 
 # As a special case, running the agent test needs /var/run/clearwater to exist on the local machine
 run_cw_alarm_test : | /var/run/clearwater
+run_cw_alarm_fvtest : | /var/run/clearwater
 /var/run/clearwater :
 	sudo mkdir -p $@
 	sudo chmod -R o+wr /var/run/clearwater
--- a/alarm_table_defs.cpp
+++ b/alarm_table_defs.cpp
@@ -69,6 +69,7 @@ bool AlarmTableDefKey::operator<(const AlarmTableDefKey& rhs) const
          ((_index == rhs._index) && (_severity < rhs._severity));
 }
 
+// LCOV_EXCL_START
 bool AlarmTableDefs::initialize(std::string& path)
 {
   std::map<unsigned int, unsigned int> dup_check;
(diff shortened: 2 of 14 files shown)


alarm_table_defs.cpp
```diff
@@ -133,14 +133,25 @@ bool AlarmTableDefs::populate_map(std::string path,
     for (s_it = a_it->_severity_details.begin(); s_it != a_it->_severity_details.end(); s_it++)
     {
       AlarmTableDef def(*a_it, *s_it);
-      AlarmTableDefKey key(a_it->_index, s_it->_severity);
-      _key_to_def.emplace(key, def);
+      AlarmTableDefs::insert_def(def);
     }
   }
-
+  AlarmDef::SeverityDetails cleared(AlarmDef::CLEARED, "Test alarm cleared description", "Test alarm cleared details");
```
rkd-msw: I think we had discussed only putting this code in `clearwater-fv-test` - as it is, this will add a line saying "Test alarm cleared description" to real servers, which I don't think is what we want. (If you think we do want that, let's discuss.)

alarm_table_defs.hpp
```diff
@@ -39,6 +39,22 @@
 
 #include "alarmdefinition.h"
 
+// Unique key for alarm table definitions is comprised of alarm index and
```
rkd-msw: Why have you moved this code up? (It makes it harder to review, as I can't tell if it's just been moved, or if you've moved it and changed the functionality too.)

alarm_table_defs.hpp
```diff
@@ -60,8 +76,6 @@ class AlarmTableDef
     _alarm_definition(alarm_definition),
     _severity_details(severity_details) {}
 
-  unsigned int state();
```
rkd-msw: Why have you removed this? I'm slightly surprised it still compiles without this line - have you run `make` and `make test` to check that?

alarm_table_defs.hpp
graemerobertson: I generally approve of keeping Rob on his toes so +1 from me

alarm_table_defs.hpp
```diff
@@ -60,8 +76,6 @@ class AlarmTableDef
...
-
```
rkd-msw: Did you mean to remove this whitespace?

alarm_table_defs.hpp
```diff
@@ -71,30 +85,12 @@ class AlarmTableDef
 
   bool is_valid() {return _valid;}
   bool is_not_clear() {return severity() != AlarmDef::CLEARED;}
-
```
rkd-msw: Did you mean to remove this whitespace? (I think it's quite useful to visually separate the private member variables.)

alarm_table_defs.hpp
```diff
@@ -116,7 +112,9 @@ class AlarmTableDefs
   // Generate alarm table definitions based on JSON files on the node
   bool initialize(std::string& path);
 
+  //AlarmTableDef create_def();
```
rkd-msw: I think you should probably just delete this commented-out declaration.

alarm_table_defs.hpp
```diff
@@ -116,7 +112,9 @@ class AlarmTableDefs
...
   // Retrieve alarm definition for specified index/severity
+  void insert_def(AlarmTableDef);
```
rkd-msw: You've added this declaration between the comment above and the function it's describing - it looks like the comment "Retrieve alarm definition for specified index/severity" describes this function, but it actually describes `get_definition`.

On that note, it would be good to have a one-line comment explaining what this function does (inserts an AlarmTableDef into the _key_to_def list, so that it can later be retrieved with get_definition).

alarms_agent.cpp
```diff
@@ -153,11 +153,9 @@ int main (int argc, char **argv)
   init_alarmModelTable();
   init_ituAlarmTable();
   init_alarmActiveTable(local_ip);
-
```
rkd-msw: More whitespace that I don't think you needed to remove.

rkd-msw:
A few comments - mostly cosmetic, but the big practical one is that I think this change would add two rows with test data to real alarm tables.

AME2:
Rob could you please have a look over this code. I was not able to query the sysUpTime oid even after including the address in fvtest.conf.
Also updated is the cpp-common pull request:
https://github.com/Metaswitch/cpp-common/pull/438

AME2:
Rob, could you please review the testing code.
One thing I've been thinking about is that I've initialised alarmActiveTable in the first alarmActive test and then just left it running throughout (making sure I clear all alarms at the end of each test). This doesn't feel optimal.
Also the index test may fail when comparing the day the alarm was raised to the current day if someone runs the test at midnight.
Also updated is cpp-common:
https://github.com/Metaswitch/cpp-common/compare/Alarms

Makefile
```diff
@@ -23,14 +23,21 @@ cw_alarm_test_SOURCES := test_main.cpp \
                          fakelogger.cpp \
                          fakezmq.cpp \
                          ${AGENT_COMMON_SOURCES}
-
+cw_alarm_fvtest_SOURCES := test_main.cpp \
+                           alarm.cpp \
+                           alarm_tables_testing.cpp\
+                           ${AGENT_COMMON_SOURCES}
+test: run_cw_alarm_test
```
rkd-msw: I don't think you need this line - `make test` should already be defined.

fvtest.conf
```diff
@@ -0,0 +1,26 @@
+agentAddress udp:16161,udp6:[::]:16161
+rocommunity clearwater
+
+view clearwater included .1
+
+# SNMP system OIDs
+view clearwater included .1.3.6.1.2.1.1.1.0
```
rkd-msw: I don't think you need most of these lines - `view clearwater included .1` should do this.

alarm_table_defs.hpp
```diff
@@ -115,6 +115,10 @@ class AlarmTableDefs
 public:
   // Generate alarm table definitions based on JSON files on the node
   bool initialize(std::string& path);
+  
+  // Insert an AlarmTableDef into the _key_to_def list, so that it cann
```
rkd-msw: Typo: cann

ut/alarm_tables_testing.cpp
```diff
@@ -0,0 +1,321 @@
...
+std::string alarmModelTable_entry = "1.3.6.1.2.1.118.1.1.2.1";
+std::string ituAlarmTable_entry = "1.3.6.1.2.1.121.1.1.1.1";
+std::string alarmActiveTable_table = "1.3.6.1.2.1.118.1.2.2";
+
+char buf[1024];
+
+TEST_F(SNMPTest, ModelTable)
+{ 
+  // Create a custom alarm definition
+  AlarmDef::SeverityDetails cleared(AlarmDef::CLEARED, "Test alarm cleared description", "Test alarm cleared details");
```
rkd-msw: I would be tempted to bring all this alarm creation code into a common SetUp function, and make defN_raised, defN_cleared, etc. into member variables. (You'd need to subclass SNMPTest to do this.) This means that it's a bit clearer what your code is actually testing - you don't start off with several lines of boilerplate just to create these.

ut/alarm_tables_testing.cpp
```diff
@@ -0,0 +1,321 @@
...
+  AlarmTableDef def_raised(example, raised);
+  AlarmTableDefs::get_instance().insert_def(def_cleared);
+  AlarmTableDefs::get_instance().insert_def(def_raised);
+
+  init_alarmModelTable();
+  
+  // Verifies the above alarm data is exactly what is stored in
+  // AlarmModelTable by querying the sixth column in alarmModelTable which
+  // should contain the description of the alarm)
+  ASSERT_STREQ("\"Test alarm cleared description\"\n", snmp_get_raw(alarmModelTable_entry + ".6.0.6666.1", buf, sizeof(buf)));
```
rkd-msw: I wonder if using more constants here might make things clearer - e.g. `alarmModelTable_entry + DESCRIPTION_COLUMN + ".0.6666" + RAISED_ALARM`.

ut/alarm_tables_testing.cpp
```diff
@@ -0,0 +1,321 @@
...
+  //details of the mapping to AlarmModelState see
+  //https://tools.ietf.org/html/rfc3877#section-5.4
+  ASSERT_STREQ("\"Test alarm raised details\"\n", snmp_get_raw(ituAlarmTable_entry + ".4.0.6666.3", buf, sizeof(buf)));
+  ASSERT_EQ(163, snmp_get(ituAlarmTable_entry + ".3.0.6666.1"));
+  ASSERT_EQ(163, snmp_get(ituAlarmTable_entry + ".3.0.6666.3"));
+}
+
+TEST_F(SNMPTest, ActiveTableMultipleAlarms)
+{
+  init_alarmActiveTable("10.154.153.133");
```
rkd-msw: You flagged that this didn't seem optimal, and you're right - for example, running just the ActiveTableClearingUnraisedAlarm test won't work.

I think there are two possibilities:

- create a new test subclass with a SetUpTestCase function that calls init_alarmActiveTable, so it will run once before any of these tests that require it. You could also have a TearDown function calling `alarmActiveTable_trap_handler(defN_cleared);` for each definition, in case any test fails to clean up its alarms.
- define a destroy_activeAlarmTable method that resets everything to a clean state, then create a new test subclass with SetUp and TearDown functions that call init_alarmActiveTable and destroy_activeAlarmTable respectively, so that each test gets a clean state.

The latter is a bit cleaner but more work.

ut/alarm_tables_testing.cpp
```diff
@@ -0,0 +1,321 @@
...
+  std::vector<std::string> entries = snmp_walk(alarmActiveTable_table);
+
+  
+  // Retrieves the current date and time.
+  struct timespec ts;
+  struct tm timing;
+  time_t rawtime;
+  clock_gettime(CLOCK_REALTIME, &ts);
+  rawtime = ts.tv_sec;
+  timing = *localtime(&rawtime);
```
rkd-msw: I think `struct tm timing = *localtime(time(NULL));` does the same thing and is more concise.

ut/alarm_tables_testing.cpp
```diff
@@ -0,0 +1,321 @@
...
+  int day = timing.tm_mday;
+  int month = 1 + timing.tm_mon;
+
+  // Splits each part of the index field (separated by ".") for the first alarm into a vector.
+  std::vector<std::string> index_fields1;
+  Utils::split_string(entries[0], '.', index_fields1);
+
+  // As defined in RFC3877 the 17th and 16th elements of the index field are the day and
```
rkd-msw: Is it clearer to split this into two tests - one checking the timestamp, the other checking that the index is increasing?

ut/alarm_tables_testing.cpp
```diff
@@ -0,0 +1,321 @@
...
+  ASSERT_EQ(std::to_string(month), index_fields1[16]);
+ 
+  // Checks the index of the first alarm contains 26 elements as defined in RFC
+  // 3877.
+  ASSERT_EQ(26, index_fields1.size());
+  
+  // Retreives each alarm's index (which should be a strictly monotonically
+  // increasing integer which indexes the entries within the table and is the
+  // 26th and last element of the index field) from the index field of the table.
+  // As the alarm's index is immediately proceeded by a space and equality
```
rkd-msw: It might be clearer to have a small helper function `std::string get_index_from_row(std::string snmpwalk_row)` which did this.

docs/Testing.md
```diff
@@ -0,0 +1,99 @@
...
+[----------] Global test environment tear-down
+[==========] 31 tests from 3 test cases ran. (316 ms total)
+[  PASSED  ] 31 tests.
+```
+
+Integration Testing
+-------------------
+
+To run the SNMP Alarm Agent integration test suite, change to the top-level
+clearwater-snmp-handlers director and issue make fvtest.
```
rkd-msw: director -> directory

docs/Testing.md
```diff
@@ -0,0 +1,99 @@
...
+
+The integration tests also use the Google Test framework and are responsible for
+testing the following components of the SNMP Alarm Agent.
+
+- The Alarm Model Table
+- The itu Alarm Table
+  - Tests that when quieried using snmp get commands these tables will output the
```
rkd-msw: quieried -> queried

rkd-msw: Also I'd say either "snmpget" (the command-line command with no space) or "SNMP GET" (the protocol and the PDU, both of which are capitalised).

docs/Testing.md
```diff
@@ -0,0 +1,99 @@
...
+    data exactly as it appears in an alarm definition. 
+- The Alarm Active Table
+  - Tests the logic behind when this table adds a row and deletes a row. For
+    example this tests that when two alarms are raises (with the second having a
```
rkd-msw: raises -> raised

docs/Testing.md
```diff
@@ -0,0 +1,99 @@
...
+    different severity) the table deletes the entry for the first alarm and
+    creates and entry for the second.
```
rkd-msw: and -> an

rkd-msw:
Comments inline, generally looked good.

(I'm not overly concerned about tests failing at midnight, but we have (cwtest_completely_control_time)[https://github.com/Metaswitch/cpp-common/blob/master/test_utils/test_interposer.cpp#L140] to deal with those kind of issues.)

One high-level question - have you run `make full_test` to get valgrind and coverage checking?

AME2:
I've applied the markups. Please could you take another look through and let me know if it's okay for merging. Includes:
https://github.com/Metaswitch/cpp-common/pull/438

Issue #116, opened 2016-01-12
Tests in clearwarer-snmp-handlers and clearwater-fvtest create a SNMP agent on the same address.


AME2:
Both clearwater-snmp-handlers/fvtest.conf and clearwater-fv-test/src/fvtest.conf register an SNMP agent on address 16161. This causes the clearwater-snmp-handlers build to fail in Jenkins as clearwarer-fv-test is built first.

To find this we should change the address of the SNMP agent within clearwater-snmp-handlers/fvtest.conf and then modify the snmp commands (snmp_get_raw and snmp_walk) to use the correct SNMP agent.

rkd-msw:
Is this now fixed?

AME2:
Yes it is.
This was fixed by including a constructor in the SNMPTest class within cpp-common that sets the address of the alarm agent that the snmp_get and snmp_walk member functions use.
Then in Clearwater_snmp_handlers in the CustomDefs class (which is a subclass of SNMPTest) its default constructor called the SNMPTest constructor to set the alarm agent to a different value (16162 as opposed to 16161 used in clearwater_fv_test).

PR #117, opened 2016-01-12
[Reviewer: Rob] Tests in clearwater-fvtest and clearwater-snmp-handlers ...

AME2:
now use different alarm agents.
This fixes issue116:
https://github.com/Metaswitch/clearwater-snmp-handlers/issues/116

### DIFF
--- a/ut/alarm_tables_testing.hpp
+++ b/ut/alarm_tables_testing.hpp
@@ -6,6 +6,8 @@
 class CustomDefs: public SNMPTest
 {
 public:
+  CustomDefs() {alarm_address = "16162";}
+  
   AlarmTableDef* def_cleared;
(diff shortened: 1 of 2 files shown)


ut/alarm_tables_testing.hpp
```diff
@@ -6,6 +6,8 @@
 class CustomDefs: public SNMPTest
 {
 public:
+  CustomDefs() {alarm_address = "16162";}
```
rkd-msw: Should you use an [initializer list](http://en.cppreference.com/w/cpp/language/initializer_list) here? We generally prefer that when a constructor is just initializing variables, as they're specifically for that.

rkd-msw:
One comment, otherwise fine to merge.

PR #123, opened 2016-02-29
[Reviewer: Andy] Move things into include/ and src/ directories

rkd-msw:
clearwater-snmp-handlers didn't have "src" and "include" directories - now it does. I've run 'make' and 'make test' successfully.


src/Makefile
```diff
@@ -0,0 +1,84 @@
...
+cdiv_handler.so_LDFLAGS := ${PLUGINS_COMMON_LDFLAGS}
+memento_as_handler.so_LDFLAGS := ${PLUGINS_COMMON_LDFLAGS}
+memento_handler.so_LDFLAGS := ${PLUGINS_COMMON_LDFLAGS}
+astaire_handler.so_LDFLAGS := ${PLUGINS_COMMON_LDFLAGS}
+
+VPATH := ../modules/cpp-common/src ../modules/cpp-common/test_utils ut
+
+GMOCK_DIR := ../modules/gmock
+GCOVR_DIR := ../modules/gcovr
+CPP_COMMON_DIR := ../modules/cpp-common
```
bossmc: From this I deduce that the gmock/gtest version in the submodules has not been updated (since the default values for these constants is correct if it has).  Not really an issue, but whatever.

src/Makefile
```diff
@@ -0,0 +1,84 @@
...
+PLUGINS_COMMON_LDFLAGS := -lzmq -lpthread `net-snmp-config --agent-libs` -fPIC -shared
+cdiv_handler.so_LDFLAGS := ${PLUGINS_COMMON_LDFLAGS}
+memento_as_handler.so_LDFLAGS := ${PLUGINS_COMMON_LDFLAGS}
+memento_handler.so_LDFLAGS := ${PLUGINS_COMMON_LDFLAGS}
+astaire_handler.so_LDFLAGS := ${PLUGINS_COMMON_LDFLAGS}
+
+VPATH := ../modules/cpp-common/src ../modules/cpp-common/test_utils ut
+
+GMOCK_DIR := ../modules/gmock
+GCOVR_DIR := ../modules/gcovr
```
bossmc: From this I deduce that the gmock/gtest version in the submodules has not been updated (since the default values for these constants is correct if it has). Not really an issue, but whatever.

src/Makefile
```diff
@@ -0,0 +1,84 @@
...
+CPP_COMMON_DIR := ../modules/cpp-common
+BUILD_DIR := ../build
```
bossmc: These, on the other hand, are the default values for these constants.

https://github.com/Metaswitch/clearwater-build-infra/blob/master/cpp.mk#L29-L30

bossmc:
Looks good, you'll need to merge up to dev and update the CI job definition to match the new targets/build artefact paths.

PR #125, opened 2016-03-02
[Reviewer: Alex] Refactors ActiveAlarmList into ObservedAlarms which keeps track of the most recent state (CLEARED or otherwise) an alarm was raised with

AME2:
I have also included a Unit Test that CLEARs an alarm without having raised it first to make sure the trap doesn't get filtered out.


alarm_trap_sender.cpp
```diff
@@ -50,35 +50,61 @@ AlarmFilter AlarmFilter::_instance;
 
 AlarmTrapSender AlarmTrapSender::_instance;
 
-bool ActiveAlarmList::update(AlarmTableDef& alarm_table_def, const std::string& issuer)
+bool ObservedAlarms::update(AlarmTableDef& alarm_table_def, const std::string& issuer)
 {
-  bool updated = false;
+  bool send_trap = false;
```
AlexHockey: I preferred the old name. The observed alarms class has no knowledge of traps - it can only say whether the observed alarms have changed, so updated was a good name for the variable.

alarm_trap_sender.cpp
```diff
@@ -50,35 +50,61 @@ AlarmFilter AlarmFilter::_instance;
...
 
   std::map<unsigned int, AlarmListEntry>::iterator it = _index_to_entry.find(alarm_table_def.alarm_index());
-
+  
+  // Enter this if statement if the alarm we are currently raising is being
+  // raised with a non-CLEARED severity. 
   if (alarm_table_def.is_not_clear())
```
AlexHockey: This entire if test could be replaced with the inner if test on line 67. Can you see why?

alarm_trap_sender.cpp
```diff
@@ -50,35 +50,61 @@ AlarmFilter AlarmFilter::_instance;
...
+      // it would be a repeated message).
     }
   }
 
-  return updated;
+  return send_trap;
 }
 
-void ActiveAlarmList::remove(ActiveAlarmIterator& it)
+void ObservedAlarms::overwrite(ObservedAlarmsIterator& it, AlarmTableDef& alarm_table_def)
```
AlexHockey: Why do you need an `overwrite` method that is different from the `update` method?

alarm_trap_sender.hpp
```diff
@@ -61,40 +61,43 @@ class AlarmListEntry
...
 
   AlarmListEntry& operator*()  {return  (std::map<unsigned int, AlarmListEntry>::iterator::operator*().second);}
   AlarmListEntry* operator->() {return &(std::map<unsigned int, AlarmListEntry>::iterator::operator*().second);}
 };
 
-// Container for all currently active alarms. Currently indexed by alarm
-// model index (may need to be extended to include a resource id going
-// forward). 
+// Container for all alarms we have seen being raised (either at CLEARED or
+// non-CLEARED severity). Maps the index of an alarms to the latest severity
```
AlexHockey: alarms -> alarm

alarm_trap_sender.hpp
```diff
@@ -61,40 +61,43 @@ class AlarmListEntry
...
 
-  // Adds a non CLEARED severity alarm to the list if it does not already
-  // exist. For a CLEARED severity update, removes an associated alarm of
-  // the same alarm model index if one exists. Returns true if a change
-  // was made to the list.
+  // Adds a non-CLEARED severity alarm to the mapping if it does not already
+  // exist and returns true. If an entry does exist but at a different severity
+  // then we update the mapping and return true. For a CLEARED severity update,
+  // if the alarm doesn't exist in the mapping or if it does exist but at a
+  // non-CLEARED severity, we update the mapping and return true.
```
AlexHockey: It sounds like the processing for Cleared and non-cleared seventies are basically the same.

ut/alarm_req_listener_test.cpp
```diff
@@ -249,8 +249,18 @@ inline Matcher<netsnmp_variable_list*> TrapVars(TrapVarsMatcher::TrapType trap_t
...
+  EXPECT_CALL(_ms, send_v2trap(TrapVars(TrapVarsMatcher::CLEAR,
+                                        1000)));
+
+  _alarm_1._clear_state.issue();
+  _ms.trap_complete(1, 5);
+}
+
 TEST_F(AlarmReqListenerTest, SetAlarm)
 {
+  advance_time_ms(AlarmFilter::ALARM_FILTER_TIME + 1);
```
AlexHockey: Why do you need this? Is it because the previous tests don't clear up after themselves? If so, could you add a comment explaining this (in addition to raising an issue - which I believe you're already doing).

AlexHockey: Is there a test where you set an alarm, wait >5s, then set it again, and check you only get one trap? If not, could you add one?

AlexHockey:
I've reviewed. As discussed, there is one key comment in the main function you've changed, and applying it should make the code a lot simpler. Please can you resend this to me when you've applied the markups?

alarm_trap_sender.cpp
AME2: Yes I do, the difference with the refactored ObservedAlarms is that it keeps track of the last state in which an alarm was raised, whether that be CLEARED or otherwise (as opposed to ActiveAlarmList which only kept track of alarms raised with a non-CLEARED severity).
That means there is no need to treat CLEARED alarms and non-CLEARED alarms differently, we just add them to the mapping if either they are not already in there, or they are but in a different state.

AlexHockey: Agreed. This also means that `update` and `override` are basically the same function, except that the latter doesn't return whether an update happened. If the caller does not care about this, it could just call `update` and ignore the return code, so we can bin `override` entirely.

ut/alarm_req_listener_test.cpp
AME2: Yes we do need to advance time to make sure we don't filter out the repeated alarms (which means we don't actually need it in the second test because there's no repeat).
If I can't fix it by tomorrow afternoon I'll add issues for both the UTs  not cleaning up and the over aggressive five second filtering.

AME2:
The main changes in the latest commit are that I've streamlined the logic to update the ObservedAlarms mapping and have added a UT to test whether if we repeat an alarm's state that a repeated trap won't get sent even if we wait for longer that the filtering time.

alarm_trap_sender.cpp
```diff
@@ -50,37 +50,28 @@ AlarmFilter AlarmFilter::_instance;
...
-      _index_to_entry[alarm_table_def.alarm_index()] = AlarmListEntry(alarm_table_def, issuer);
-      updated = true;
-    }
-  }
-  else
+  
+  // If either the current alarm doesn't exist in the ObservedAlarms mapping
+  // or there is an entry for the current alarm in the mapping but at a
+  // different severity to the one we are currently raising the alarm with, we
+  // want to update the mapping and send a trap for the alarm.
```
AlexHockey: Please could you remove the reference to sending a trap - as I said before, this class doesn't know about traps.

alarm_trap_sender.cpp
```diff
@@ -50,37 +50,28 @@ AlarmFilter AlarmFilter::_instance;
...
-    {
-      _index_to_entry.erase(it);
-      updated = true;
-    }
+    _index_to_entry[alarm_table_def.alarm_index()] = AlarmListEntry(alarm_table_def, issuer);
+    updated = true;
   }
+  // If the alarm we are currently trying to raise has an associated entry
+  // within the ObservedAlarms mapping at the same severity then we do not
+  // wish to send a trap for this alarm (as it would be a repeated message).
```
AlexHockey: I don't think this comment is needed.

ut/alarm_req_listener_test.cpp
```diff
@@ -258,8 +267,14 @@ TEST_F(AlarmReqListenerTest, SetAlarm)
   _ms.trap_complete(1, 5);
 }
 
+// We have to advance time here so that our alarms are not filtered out by the
+// alarm_filtered function with the alarm trap sender. This filters out any
+// alarms raised in a repeated state during five seconds of each other (the
+// value of ALARM_FILTER_TIME), even if the state of the alarm changes with that
+// five seconds.
```
AlexHockey: The key point is that we are using the same `AlarmTrapSender` for all the tests, which means this test doesn't start from a clean state, so we have to wait for the filters to expire before doing our test.

ut/alarm_req_listener_test.cpp
```diff
@@ -269,6 +284,28 @@ TEST_F(AlarmReqListenerTest, ClearAlarm)
   _ms.trap_complete(1, 5);
 }
 
+// Raises an alarm, waits thirty seconds, raises the same alarm in the
+// same state and then clears the alarm. We should only expect two traps, for
+// the initial raising and the clearing, raising the alarm in a repeated state
+// should not cause a trap to be sent.
```
AlexHockey: Ditto

AlexHockey:
Code looks good, I just have a few bits of feedback regarding commenting. You can merge once you've fixed these up. 

One more thing - have we tested that the active alarms table is still working? I think we should so if it isn't tested elsewhere, can you think about how best to test it? The particular case I'm worried about is that cleared alarms shouldn't be listed when iterating over the table.

AME2:
Yes I've live tested the Active Alarm Table and the old UTs for all alarm tables still pass.
The reason why cleared alarms would never be displayed in the Active Alarm Table is that the trap handler within the table only uses CLEARED alarms to delete entries in the table, never add them.
https://github.com/Metaswitch/clearwater-snmp-handlers/blob/master/alarm_active_table.cpp#L314

PR #129, opened 2016-03-15
[Reviewer: Rob] Remove clear all alarms

johadalin:
As we discussed, the trap sender now sends all alarms it knows about, set or cleared. 

As there is no longer a clear_all method for alarms, I am removing the UT for it.


rkd-msw:
Yes.

PR #131, opened 2016-03-18
Reviewer Rob - Use SNMP NOTIFY callbacks to re-send unacknowledged alarms

bossmc:
Registers for callbacks when sending SNMP alarms and, if the send failed **and** the alarm is still active, resend the TRAP.

The only slight nasty is that I have to clone the `AlarmTableDef` object to use as the alarm correlator, which then gets cleaned up after the callback pops.  You'll see in the UTs I've added a utility function that sorts this out in the tests if all they need is for the alarm to succeed.

### DIFF
--- a/alarm_trap_sender.cpp
+++ b/alarm_trap_sender.cpp
@@ -52,15 +52,13 @@ AlarmTrapSender AlarmTrapSender::_instance;
 
 bool ObservedAlarms::update(const AlarmTableDef& alarm_table_def, const std::string& issuer)
 {
+  bool updated = false;
   if (!is_active(alarm_table_def))
   {
     _index_to_entry[alarm_table_def.alarm_index()] = AlarmListEntry(alarm_table_def, issuer);
-    return true;
+    updated = true;
   }
-  else
-  {
-    return false;
-  }
+  return updated;
 }
 
 bool ObservedAlarms::is_active(const AlarmTableDef& alarm_table_def)
@@ -72,12 +70,15 @@ bool ObservedAlarms::is_active(const AlarmTableDef& alarm_table_def)
     // Either the current alarm doesn't exist in the ObservedAlarms mapping
     // or there is an entry for the current alarm in the mapping but at a
     // different severity to the one we are currently raising the alarm with.
-    TRC_DEBUG("Alarm is not active at this severity");
+    TRC_DEBUG("Alarm is active at %d severity which does not match the given severity (%d)",
+              it->second.alarm_table_def().severity(),
+              alarm_table_def.severity());
     return false;
   }
   else
   {
-    TRC_DEBUG("Alarm is active at this severity");
+    TRC_DEBUG("Alarm is active at the given severity (%d)",
+              alarm_table_def.severity());
     return true;
   }
 }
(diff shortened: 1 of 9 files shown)


alarm_trap_sender.cpp
```diff
@@ -182,11 +196,52 @@ void AlarmTrapSender::sync_alarms(bool do_clear)
...
+                                               const AlarmTableDef& alarm_table_def)
+{
+  switch (op)
+  {
+  case NETSNMP_CALLBACK_OP_RECEIVED_MESSAGE:
+    TRC_DEBUG("Alarm successfully delivered");
+    break;
+  case NETSNMP_CALLBACK_OP_SEND_FAILED:
+    // This can happen pretty fast, so don't spin
+    sleep(1);
```
rkd-msw: What thread are you blocking here? I suspect it's our sole netsnmp thread, and so this sleep stops us responding to SNMP requests.

alarm_trap_sender.cpp
```diff
@@ -50,22 +50,36 @@ AlarmFilter AlarmFilter::_instance;
...
-  // want to update the mapping.
+
   if ((it == _index_to_entry.end()) || (alarm_table_def.severity() != it->second.alarm_table_def().severity()))
   {
-    _index_to_entry[alarm_table_def.alarm_index()] = AlarmListEntry(alarm_table_def, issuer);
-    updated = true;
+    // Either the current alarm doesn't exist in the ObservedAlarms mapping
+    // or there is an entry for the current alarm in the mapping but at a
+    // different severity to the one we are currently raising the alarm with.
+    TRC_DEBUG("Alarm is not active at this severity");
```
rkd-msw: If it's active at another severity, might be worth TRC_DEBUGing that severity out.

alarm_trap_sender.cpp
```diff
@@ -50,22 +50,36 @@ AlarmFilter AlarmFilter::_instance;
...
 AlarmTrapSender AlarmTrapSender::_instance;
 
-bool ObservedAlarms::update(AlarmTableDef& alarm_table_def, const std::string& issuer)
+bool ObservedAlarms::update(const AlarmTableDef& alarm_table_def, const std::string& issuer)
 {
-  bool updated = false;
+  if (!is_active(alarm_table_def))
+  {
+    _index_to_entry[alarm_table_def.alarm_index()] = AlarmListEntry(alarm_table_def, issuer);
+    return true;
```
rkd-msw: I quite liked the `updated = true; ... return updated` because it was obvious what the return value meant.

alarm_trap_sender.cpp
```diff
@@ -182,11 +196,52 @@ void AlarmTrapSender::sync_alarms(bool do_clear)
...
+    TRC_DEBUG("Failed to deliver alarm");
+    if (_observed_alarms.is_active(alarm_table_def))
+    {
+      // Alarm is still active, attempt to re-transmit
+      send_trap(alarm_table_def);
+    }
+    break;
+  default:
+    // LCOV_EXCL_START - Logical error
+    TRC_DEBUG("Ignoring uninteresting alarm delivery callback (%d)", op);
```
rkd-msw: Uninteresting, or unexpected? Presumably the latter if this is a logical error.

rkd-msw:
My main comment here was that I'd expected a different structure - a data structure tracking whether alarms were acknowledged/unacknowledged/in-flight, and a thread which checked that data structure periodically and retransmitted alarms.

I think that design would make it easier to do pacing - here, you can only limit the rate of alarm retransmission by blocking our SNMP thread.

It's also probably a bit better diagnostically, because you have the current state of all alarms to inspect if you want.

alarm_trap_sender.cpp
```diff
@@ -182,11 +196,52 @@ void AlarmTrapSender::sync_alarms(bool do_clear)
...
+  case NETSNMP_CALLBACK_OP_SEND_FAILED:
+    // This can happen pretty fast, so don't spin
+    sleep(1);
+    // Fall through to...
+  case NETSNMP_CALLBACK_OP_TIMED_OUT:
+    TRC_DEBUG("Failed to deliver alarm");
+    if (_observed_alarms.is_active(alarm_table_def))
+    {
+      // Alarm is still active, attempt to re-transmit
+      send_trap(alarm_table_def);
```
rkd-msw: This retransmits to all configured alarm destinations, even if only one is down, right?

alarm_trap_sender.cpp
```diff
@@ -182,11 +196,52 @@ void AlarmTrapSender::sync_alarms(bool do_clear)
   }
 }
 
+static int alarm_trap_send_callback(int op,
+                                    snmp_session* session,
```
rkd-msw: Maybe worth pulling out session->peername for diagnostic purposes?

alarm_trap_sender.cpp
```diff
@@ -182,11 +197,60 @@ void AlarmTrapSender::sync_alarms(bool do_clear)
...
+  case NETSNMP_CALLBACK_OP_RECEIVED_MESSAGE:
+    TRC_DEBUG("Alarm successfully delivered to %s", peer.c_str());
+    break;
+  case NETSNMP_CALLBACK_OP_SEND_FAILED:
+    // There is no path through NETSNMP that uses this OP.  If there were, we'd
+    // want to delay before re-trying, but this is non-trivial to accomplish
+    // with the current architecture.  For now, if this OP occurs, we'll treat
+    // it as a successful send.
+    // LCOV_EXCL_START - Unhittable
+    TRC_DEBUG("Ignoring failed alarm send to peer %s", peer.c_str());
```
rkd-msw: TRC_WARN? I'm worried we have a possible logic bug we won't spot.

alarm_trap_sender.cpp
```diff
@@ -68,6 +61,28 @@ bool ObservedAlarms::update(AlarmTableDef& alarm_table_def, const std::string& i
...
+{
+  std::map<unsigned int, AlarmListEntry>::iterator it = _index_to_entry.find(alarm_table_def.alarm_index());
+
+  if ((it == _index_to_entry.end()) || (alarm_table_def.severity() != it->second.alarm_table_def().severity()))
+  {
+    // Either the current alarm doesn't exist in the ObservedAlarms mapping
+    // or there is an entry for the current alarm in the mapping but at a
+    // different severity to the one we are currently raising the alarm with.
+    TRC_DEBUG("Alarm is active at %d severity which does not match the given severity (%d)",
+              it->second.alarm_table_def().severity(),
```
rkd-msw: Won't you crash here if `it == _index_to_entry.end()`?

rkd-msw:
Two comments, then ship it. I think you're also raising a story to refactor and get a single point of truth for alarm state, following my design comment?

alarm_trap_sender.cpp
bossmc: Yes... I'm surprised my test ran...

PR #136, opened 2016-04-13
[Reviewer: Andrew] Fix up warnings when compiling with clang

rkd-msw:
This fixes up a couple of compiler warnings I saw when compiling with a newer compiler (Clang 3.8):

* preferring std::abs to abs when getting absolute values (http://stackoverflow.com/questions/21392627/abs-vs-stdabs-what-does-the-reference-say)
* explicitly marking integer literals as unsigned, if we're comparing them to unsigned values

### DIFF
--- a/alarm_active_table.cpp
+++ b/alarm_active_table.cpp
@@ -289,7 +289,7 @@ alarmActiveTable_SNMPDateTime alarm_time_issued(void)
   // tells us how offset (in seconds) from UTC we are and stores a 
   // broken down time representation in the structure timing
   timing = *localtime(&rawtime);
-  UTC_offset = abs(timezone) / 3600;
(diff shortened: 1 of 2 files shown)


AME2:
The very last comparison in ut/alarm_tables_testing.cpp:
ASSERT_LT(atoi(get_index_from_row(entries[0]).c_str()), atoi(get_index_from_row(entries[13]).c_str()));
hasn't been explicitly marked as unsigned.
Other than this it all looks good.

rkd-msw:
That's because it isn't a comparison between signed and unsigned numbers, so doesn't (and shouldn't) cause a compiler waning - atoi returns an int, so it's a comparison between two signed numbers.

PR #137, opened 2016-04-13
[Reviewer: Andrew] Allow running UT without /var/run/clearwater

rkd-msw:
Previously, these UTs had to be run as root (at least once) because they needed to create /var/run/clearwater. This changes the alarm location we use in UT to a file in /tmp, so that it doesn't need to run as root.


AME2:
Looks good

PR #147, opened 2016-05-26
[Reviewer: Rob] Alarm refactor

eleanor-merry:
Rob, can you please review this refactoring of the alarm agent.

I've changed how INFORMs are sent by the alarm agent, splitting up the sending INFORMs into its own component, and adding a heap that alarms are queued on (to make the retry/resync/flicker behaviour correct and easy to maintain).

I've also moved us away from using singletons


src/alarm_heap.cpp
```diff
@@ -0,0 +1,326 @@
...
+
+    AllAlarmsStateMap::iterator alarm = _all_alarms_state.find(index);
+
+    if (alarm != _all_alarms_state.end())
+    {
+      AlarmDef::Severity severity_as_enum = (AlarmDef::Severity)severity;
+      AlarmSeverityChange severity_change =
+                  (*alarm).second->new_alarm_severity(severity_as_enum);
+
+      if (severity_change == AlarmSeverityChange::INCREASED_SEVERITY)
```
rkd-msw: Could you define a custom `operator >` for AlarmDef::Severity, so that this could be `if (severity > old_severity)`? That feels more natural to read - you don't have to understand (or have) `new_alarm_severity` or `AlarmSeverityChange`.

include/alarm_heap.hpp
```diff
@@ -0,0 +1,198 @@
...
+    _severity(severity)
+  {}
+
+  ~AlarmInfo();
+
+  // Compares a new severity with the current severity for this alarm,
+  // and returns the severity change
+  // @params severity - the new severity
+  // @returns         - the severity change
+  AlarmSeverityChange new_alarm_severity(AlarmDef::Severity severity);
```
rkd-msw: If you keep this (see https://github.com/Metaswitch/clearwater-snmp-handlers/pull/147/files#r64817975), can it be called `compare_alarm_severity`?

include/alarm_heap.hpp
```diff
@@ -0,0 +1,198 @@
...
+
+  // Updates the alarm in the heap's severity, and its time to pop
+  void update_alarm_in_heap(AlarmDef::Severity new_severity,
+                            uint64_t time_to_delay);
+
+  AlarmInHeap* alarm_in_heap() { return _alarm_in_heap; }
+  AlarmDef::Severity get_severity() { return _severity; }
+  void set_severity(AlarmTableDef& alarm_table_def);
```
rkd-msw: For symmetry, should this take an AlarmDef::Severity?

include/alarm_heap.hpp
```diff
@@ -0,0 +1,198 @@
...
+
+enum AlarmSeverityChange
+{
+  NO_CHANGE,
+  INCREASED_SEVERITY,
+  REDUCED_SEVERITY,
+};
+
+
+class AlarmInHeap : public HeapableTimer
```
rkd-msw: Could you call this AlarmTimer? It's a bit picky, but an AlarmInHeap object currently has an in_heap method that can return false, which feels odd.

include/alarm_heap.hpp
```diff
@@ -0,0 +1,198 @@
...
+  ~AlarmInfo();
+
+  // Compares a new severity with the current severity for this alarm,
+  // and returns the severity change
+  // @params severity - the new severity
+  // @returns         - the severity change
+  AlarmSeverityChange new_alarm_severity(AlarmDef::Severity severity);
+
+  // Updates the alarm in the heap's severity, and its time to pop
+  void update_alarm_in_heap(AlarmDef::Severity new_severity,
```
rkd-msw: Is this clearer if it's called `schedule_change` or something?

include/alarm_heap.hpp
```diff
@@ -0,0 +1,198 @@
...
+                            uint64_t time_to_delay);
```
rkd-msw: In what units?

include/alarm_heap.hpp
```diff
@@ -0,0 +1,198 @@
...
+    // the severity.
+    return ((_severity == severity) && (!_alarm_in_heap->in_heap()));
+  }
+
+private:
+  AlarmInHeap* _alarm_in_heap;
+  AlarmDef::Severity _severity;
+};
+
+typedef std::map<unsigned int, AlarmInfo*> AllAlarmsStateMap;
```
rkd-msw: I'm not sure this typedef makes things much clearer. (Pedantically, the fact that this contains *all* alarms isn't type system information.)

A typedef that might be good is `typedef unsigned int AlarmIndex`. `std::map<AlarmIndex, unsigned int>& dup_check` is clearer than `std::map<unsigned int, unsigned int>& dup_check`, for example.

include/alarm_heap.hpp
```diff
@@ -0,0 +1,198 @@
...
+  AlarmHeap(AlarmTableDefs* alarm_table_defs);
+  virtual ~AlarmHeap();
+
+  // Generates an alarmActiveState inform if the identified alarm is not
+  // of CLEARED severity and not already active (or subject to filtering).
+  // Generates an alarmClearState inform if the identified alarm is of a
+  // CLEARED severity and an associated alarm is active (unless subject
+  // to filtering).
+  virtual void issue_alarm(const std::string& issuer,
+                   const std::string& identifier);
```
rkd-msw: Indentation.

include/alarm_trap_sender.hpp
```diff
@@ -43,145 +43,44 @@
...
-
-  std::map<AlarmFilterKey, unsigned long> _issue_times;
-
-  unsigned long _clean_time;
-
-  static AlarmFilter _instance;
-};
+class AlarmHeap;
 
 // Singleton class providing methods for generating alarmActiveState and
```
rkd-msw: Singleton?

include/alarm_trap_sender.hpp
```diff
@@ -43,145 +43,44 @@
...
-// alarmClearState inform notifications. It maintains an observed alarm 
-// list to support filtering and synchronization. 
-
+// alarmClearState inform notifications. It maintains an observed alarm
```
rkd-msw: Out-of-date comment?

src/alarm_heap.cpp
```diff
@@ -0,0 +1,326 @@
...
+
+  while (!_terminated)
+  {
+    uint64_t time_now_in_ms = Utils::get_time();
+    AlarmInHeap* alarm_in_heap = (AlarmInHeap*)_alarm_heap.get_next_timer();
+
+    while ((alarm_in_heap) &&
+           (alarm_in_heap->get_pop_time() <= time_now_in_ms))
+    {
+      // While the next alarm in the heap is due to be sent, we:
```
rkd-msw: "For each alarm in the heap that's due to be sent..."? A bit clearer that you're not just talking about one alarm.

src/alarm_heap.cpp
```diff
@@ -0,0 +1,326 @@
...
+  pthread_join(_heap_sender_thread, NULL);
+  delete _cond; _cond = NULL;
+
+  pthread_mutex_destroy(&_lock);
+
+  delete _alarm_trap_sender; _alarm_trap_sender = NULL;
+  _alarm_heap.clear();
+
+  for (AllAlarmsStateMap::iterator it = _all_alarms_state.begin();
+                               it != _all_alarms_state.end();
```
rkd-msw: Indentation.

src/alarm_trap_sender.cpp
```diff
@@ -43,167 +43,24 @@
...
 static int alarm_trap_send_callback(int op,
                                     snmp_session* session,
                                     int req_id,
                                     snmp_pdu* pdu,
                                     void* correlator)
 {
-  AlarmTableDef* alarm_table_def = (AlarmTableDef *)correlator; correlator = NULL;
-  AlarmTrapSender::get_instance().alarm_trap_send_callback(op,
-                                                           *alarm_table_def);
+  ((SNMPCallbackAlarmInfo*)correlator)->_ats->alarm_trap_send_callback(
```
rkd-msw: Clearer as two lines?

```
SNMPCallbackAlarmInfo* cbi = static_cast<SNMPCallbackAlarmInfo*>(correlator);
cbi->_ats->alarm_trap_send_callback(op, cbi->_atd);
```

Also, _ats and _atd are a bit obscure - `_sender` and `_definition`?

src/alarm_trap_sender.cpp
```diff
@@ -241,7 +96,7 @@ void AlarmTrapSender::alarm_trap_send_callback(int op,
 // retires if needed.
 void AlarmTrapSender::send_trap(const AlarmTableDef& alarm_table_def)
 {
-  TRC_WARNING("Trap with alarm ID %d.%d being sent", alarm_table_def.alarm_index(), alarm_table_def.state());
+  TRC_INFO("Trap with alarm ID %d.%d being sent", alarm_table_def.alarm_index(), alarm_table_def.state());
```
rkd-msw: I kind of liked it as a warning because traps usually mean something's wrong, and it's nice to have a logfile recording what traps went out when.

src/alarm_trap_sender.cpp
```diff
@@ -284,8 +139,9 @@ void AlarmTrapSender::send_trap(const AlarmTableDef& alarm_table_def)
   snmp_set_var_objid(&var_resource_id, resource_id_oid, OID_LENGTH(resource_id_oid));
   snmp_set_var_typed_value(&var_resource_id, ASN_OBJECT_ID, (u_char*) zero_dot_zero, sizeof(zero_dot_zero));
 
-  send_v2trap(&var_trap, ::alarm_trap_send_callback, (void*)&alarm_table_def);
+  SNMPCallbackAlarmInfo* cbi = new SNMPCallbackAlarmInfo(this, (AlarmTableDef&)alarm_table_def);
```
rkd-msw: Where do you delete this?

src/alarm_trap_sender.cpp
```diff
@@ -43,167 +43,24 @@
...
-                                                           *alarm_table_def);
+  ((SNMPCallbackAlarmInfo*)correlator)->_ats->alarm_trap_send_callback(
+                                op, ((SNMPCallbackAlarmInfo*)correlator)->_atd);
   return 1;
 }
 
-void AlarmTrapSender::alarm_trap_send_callback(int op,
-                                               const AlarmTableDef& alarm_table_def)
+void AlarmTrapSender::alarm_trap_send_callback(
+                                           int op,
```
rkd-msw: Indentation.

include/alarm_heap.hpp
rkd-msw: Also maybe worth a class comment to say what it's for.

include/alarm_heap.hpp
```diff
@@ -0,0 +1,198 @@
...
+  bool in_heap() { return (_heap != nullptr); }
+  uint64_t get_pop_time() const { return _pop_time; }
+
+private:
+  uint64_t _pop_time;
+  AlarmDef::Severity _severity;
+  unsigned int _index;
+};
+
+class AlarmInfo
```
rkd-msw: How are AlarmTableDef, AlarmInHeap and AlarmInfo distinct from each other? At the very least, we should have a comment above each explaining its role (and maybe better names?).

include/alarm_heap.hpp
```diff
@@ -0,0 +1,198 @@
...
+// This class accepts alarm triggers, checks that they correspond to valid
+// alarms, and passes valid alarms to the alarm trap sender to send as SNMP
+// INFORMs. It also filters the alarms.
+class AlarmHeap
+{
+public:
+  // Times to send an alarms. If we're reducing the severity of an alarm we
+  // add a delay of 30secs (to protect against flicker). Increased severity
+  // alarms or resyncing alarms are sent immediately.
+  const static uint64_t ALARM_REDUCED = 30000;
```
rkd-msw: Why do some of these constants end in _DELAY but not others?

src/ut/alarm_heap_test.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+class SNMPCallbackCollector
+{
+public:
+  void call_all_callbacks()
+  {
+    snmp_session session;
+    session.peername = strdup("peer");
+
+    for (auto ii = _callbacks.begin();
```
rkd-msw: 'auto'

src/ut/alarm_heap_test.cpp
```diff
@@ -0,0 +1,362 @@
...
+using ::testing::ReturnNull;
+using ::testing::InSequence;
+using ::testing::MakeMatcher;
+using ::testing::Matcher;
+using ::testing::MatcherInterface;
+using ::testing::MatchResultListener;
+using ::testing::SaveArg;
+using ::testing::Invoke;
+
+class TrapVarsMatcher : public MatcherInterface<netsnmp_variable_list*> {
```
rkd-msw: { should be on next line.

src/ut/alarm_heap_test.cpp
rkd-msw: Ah, this is just moved from alarm_req_listener_test.cpp.

src/ut/alarm_tables_testing.cpp
```diff
@@ -63,10 +63,11 @@ char buf[1024];
 
 TEST_F(CustomDefs, ModelTable)
 { 
-  AlarmTableDefs::get_instance().insert_def(*def_cleared);
-  AlarmTableDefs::get_instance().insert_def(*def_raised);
+  AlarmTableDefs* alarm_table_defs = new AlarmTableDefs();
```
rkd-msw: Why not just `AlarmTableDefs alarm_table_defs`? It doesn't look like it needs to be on the heap and that saves you a line of cleanup code.

src/alarm_trap_sender.cpp
```diff
@@ -220,13 +77,11 @@ void AlarmTrapSender::alarm_trap_send_callback(int op,
     break;
     // LCOV_EXCL_STOP
   case NETSNMP_CALLBACK_OP_TIMED_OUT:
-    TRC_DEBUG("Failed to deliver alarm");
-    if (_observed_alarms.is_active(alarm_table_def))
     {
```
rkd-msw: Looks a bit odd to only have braces for this case statement.

src/alarm_heap.cpp
```diff
@@ -0,0 +1,326 @@
...
+            alarm_table_def.alarm_index());
+
+  pthread_mutex_lock(&_lock);
+
+  AllAlarmsStateMap::iterator alarm =
+                          _all_alarms_state.find(alarm_table_def.alarm_index());
+
+  if (alarm != _all_alarms_state.end())
+  {
+    if ((*alarm).second->should_resend_alarm(alarm_table_def.severity()))
```
rkd-msw: `alarm->second`, surely?

rkd-msw: Also elsewhere, e.g. it->second above.

src/alarm_req_listener.cpp
```diff
@@ -238,11 +230,11 @@ void AlarmReqListener::listener()
     // section 5.4: https://tools.ietf.org/html/rfc3877#section-5.4
     if ((msg[0].compare("issue-alarm") == 0) && (msg.size() == 3))
     {
-      AlarmTrapSender::get_instance().issue_alarm(msg[1], msg[2]);
+      _alarm_heap->issue_alarm(msg[1], msg[2]);
```
rkd-msw: Hmm...conceptually the alarm heap *doesn't* issue alarms. Should this be "schedule_alarm"?

src/alarm_req_listener.cpp
```diff
@@ -238,11 +230,11 @@ void AlarmReqListener::listener()
...
     }
     else if ((msg[0].compare("sync-alarms") == 0) && (msg.size() == 1))
     {
-      AlarmTrapSender::get_instance().sync_alarms();
+      _alarm_heap->sync_alarms();
```
rkd-msw: Not sure about this one - `schedule_all_alarms_immediately()`?

src/alarm_heap.cpp
```diff
@@ -0,0 +1,326 @@
...
+      if (severity_change == AlarmSeverityChange::INCREASED_SEVERITY)
+      {
+        TRC_DEBUG("Severity of the alarm %lu has increased", index);
+        update_alarm_in_heap((*alarm).second, severity_as_enum, ALARM_INCREASED);
+        _cond->signal();
+      }
+      else if (severity_change == AlarmSeverityChange::REDUCED_SEVERITY)
+      {
+        TRC_DEBUG("Severity of the alarm %lu has reduced", index);
+        _alarm_heap.insert((*alarm).second->alarm_in_heap());
```
rkd-msw: Why do we insert for the reduced case but not for the increased case?

src/alarm_heap.cpp
```diff
@@ -0,0 +1,326 @@
...
+  {
+    AllAlarmsStateMap::iterator index = _all_alarms_state.find(it->alarm_index());
+
+    if (index == _all_alarms_state.end())
+    {
+      // We only have one entry for each alarm index (not one per severity).
+      AlarmInHeap* alarm_in_heap = new AlarmInHeap(it->alarm_index());
+      AlarmInfo* alarm_info = new AlarmInfo(alarm_in_heap,
+                                            AlarmDef::Severity::UNDEFINED_SEVERITY);
+      _all_alarms_state.insert(std::pair<unsigned int, AlarmInfo*>(it->alarm_index(),
```
rkd-msw: `std::make_pair(it->alarm_index(), alarm_info)`?

src/alarm_heap.cpp
```diff
@@ -0,0 +1,326 @@
...
+      AlarmTableDef& alarm_table_def =
+              _alarm_table_defs->get_definition(alarm_in_heap->index(),
+                                                alarm_in_heap->severity());
+
+      if (alarm_table_def.is_valid())
+      {
+        _alarm_trap_sender->send_trap(alarm_table_def);
+        AllAlarmsStateMap::iterator alarm = _all_alarms_state.find(alarm_in_heap->index());
+
+        if (alarm != _all_alarms_state.end())
```
rkd-msw: You should never hit the case where the alarm table def is valid but the index isn't in _all_alarms_state, right? Worth an error log if we do?

src/alarm_heap.cpp
```diff
@@ -0,0 +1,326 @@
...
+
+  for (AllAlarmsStateMap::iterator it = _all_alarms_state.begin();
+                               it != _all_alarms_state.end();
+                               it++)
+  {
+    AlarmDef::Severity current_severity = (*it).second->get_severity();
+
+    if (current_severity != AlarmDef::Severity::UNDEFINED_SEVERITY)
+    {
+      update_alarm_in_heap((*it).second, current_severity, ALARM_RESYNC_DELAY);
```
rkd-msw: Hmm. What if:

* I raise alarm A - it goes out
* I clear alarm A - we schedule a timer to send that in 30s
* Before the 30s are up, I synchronise alarms

I *think* current_severity is still 'raised' at that point, so this will overwrite the pending cleared state.

src/alarm_heap.cpp
```diff
@@ -0,0 +1,326 @@
...
+  }
+}
+
+void AlarmHeap::sync_alarms()
+{
+  TRC_STATUS("Resyncing all alarms");
+
+  pthread_mutex_lock(&_lock);
+
+  for (AllAlarmsStateMap::iterator it = _all_alarms_state.begin();
```
rkd-msw: I've started preferring range-based for loops as they're a bit more readable:

```
for (std::pair<unsigned int, AlarmInfo*> pair : _all_alarms_state)
{

}
```

include/alarm_heap.hpp
```diff
@@ -0,0 +1,198 @@
...
+  // of CLEARED severity and not already active (or subject to filtering).
+  // Generates an alarmClearState inform if the identified alarm is of a
+  // CLEARED severity and an associated alarm is active (unless subject
+  // to filtering).
+  virtual void issue_alarm(const std::string& issuer,
+                   const std::string& identifier);
+
+  static void* heap_sender_function(void* data);
+
+  void heap_sender();
```
rkd-msw: I was going to complain that this wasn't a very good name, but the best i could come up with was `popping_loop`.

src/alarm_heap.cpp
```diff
@@ -0,0 +1,326 @@
...
+        TRC_DEBUG("Severity of the alarm %lu has reduced", index);
+        _alarm_heap.insert((*alarm).second->alarm_in_heap());
+        update_alarm_in_heap((*alarm).second, severity_as_enum, ALARM_REDUCED);
+        _cond->signal();
+      }
+      else
+      {
+        // The severity of the alarm hasn't changed. Remove it from the heap
+        // (safe to call even if it's already not in the heap).
+        _alarm_heap.remove((*alarm).second->alarm_in_heap());
```
rkd-msw: Why's it in the heap?

rkd-msw: (The only case I can think of is if the alarm send failed, and it's gone back into the heap - at the same severity - for a retry. In this case a new ZMQ message for that alarm at the same severity would cancel the retry, I think, which is wrong.)

include/alarm_table_defs.hpp
```diff
@@ -115,35 +115,32 @@ class AlarmTableDefsIterator : public std::map<AlarmTableDefKey, AlarmTableDef>:
 class AlarmTableDefs
 {
 public:
+  // Constructor
+  AlarmTableDefs() {}
```
rkd-msw: If it's just the default constructor, you don't need it.

src/alarms_agent.cpp
```diff
@@ -139,28 +140,38 @@ int main (int argc, char **argv)
...
+  init_ituAlarmTable(*alarm_table_defs);
   init_alarmActiveTable(local_ip);
  
   init_snmp_handler_threads("clearwater-alarms");
 
+  // Exit if the ReqListener wasn't able to fully start
+  if (!alarm_req_listener->start(&term_sem))
+  {
+    TRC_ERROR("Hit error starting the listener - shutting down");
+    return 0;
```
rkd-msw: Shouldn't we return something non-zero in error cases?

src/ut/alarm_table_defs_test.cpp
```diff
@@ -53,44 +53,45 @@ class AlarmTableDefsTest : public ::testing::Test
...
 
   virtual ~AlarmTableDefsTest()
   {
     cwtest_restore_netsnmp();
+    delete _defs; _defs = NULL;
   }
 
 private:
-  AlarmTableDefs& _defs;
+  AlarmTableDefs* _defs;
```
rkd-msw: Same question as https://github.com/Metaswitch/clearwater-snmp-handlers/pull/147/files#r64824855.

include/alarm_heap.hpp
rkd-msw: Actually, thinking about file layout - should we group all the classes that somehow represent a single alarm (AlarmTableDef, AlarmInHeap and AlarmInfo) into one file, and all the classes that are alarm containers (AlarmTableDefs, AlarmHeap) into another?

include/alarm_heap.hpp
```diff
@@ -0,0 +1,198 @@
...
+  void update_alarm_in_heap(AlarmInfo* alarm_info,
+                            AlarmDef::Severity severity,
+                            uint64_t alarm_pop_time);
+
+  AlarmTableDefs* _alarm_table_defs;
+  AlarmTrapSender* _alarm_trap_sender;
+  TimerHeap _alarm_heap;
+
+  // Map holding all alarms. This maps the index of an alarm to information
+  // about the alarm (its current severity and its representation in the heap).
```
rkd-msw: Can we be a bit more precise than "current severity"? I think it's specifically "the severity we last tried to send to the NMS" (not "the severity we last successfully sent to the NMS" or "the severity we were last told about").

src/ut/alarm_heap_test.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+  COLLECT_CALL(send_v2trap(TrapVars(TrapVarsMatcher::CLEAR,
+                                        1002), _, _));
+
+  _alarm_heap->sync_alarms();
+
+  _ms.trap_complete(3, 5);
+}
+
+// Test that an alarm flicker situation doesn't cause flickering INFORMs
```
rkd-msw: I can't immediately see what about this test would fail if it did flicker - worth a clarifying comment?

src/ut/alarm_heap_test.cpp
```diff
@@ -0,0 +1,362 @@
...
+                   SaveArg<2>(&correlator)));
+  _alarm_heap->issue_alarm("test", "1000.3");
+  _ms.trap_complete(1, 5);
+
+  // Now report the send as failed.
+  snmp_session session;
+  session.peername = strdup("peer");
+  callback(NETSNMP_CALLBACK_OP_TIMED_OUT, &session, 2, NULL, correlator);
+  free(session.peername);
+  delete (SNMPCallbackAlarmInfo*)correlator;
```
rkd-msw: I think having this line in your UT is masking a leak in your production code.

src/ut/alarm_heap_test.cpp
```diff
@@ -0,0 +1,362 @@
...
+  // Now advance time by the retry delay amount. This triggers the retry to be
+  // sent
+  COLLECT_CALL(send_v2trap(TrapVars(TrapVarsMatcher::ACTIVE,
+                                    1000), _, _));
+  cwtest_advance_time_ms(AlarmHeap::ALARM_RETRY_DELAY);
+  _alarm_heap->_cond->signal();
+  _ms.trap_complete(1, 5);
+}
+
+// Test that when the raise alarm fails, but a clear alarm has been sent it later,
```
rkd-msw: sent it -> sent in?

src/ut/alarm_heap_test.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+  COLLECT_CALL(send_v2trap(TrapVars(TrapVarsMatcher::CLEAR,
+                                    1000), _, _));
+  cwtest_advance_time_ms(AlarmHeap::ALARM_REDUCED);
+  _alarm_heap->_cond->signal();
+  _ms.trap_complete(1, 5);
+}
+
+// Test handling of an invalid alarm
+TEST_F(AlarmHeapTest, InvalidAlarmIdentifier)
```
rkd-msw: Should it be the alarm heap's responsibility to do this validation?

src/alarm_heap.cpp
```diff
@@ -0,0 +1,326 @@
...
+* respects for all of the code used other than OpenSSL.
+* "OpenSSL" means OpenSSL toolkit software distributed by the OpenSSL
+* Project and licensed under the OpenSSL Licenses, or a work based on such
+* software and licensed under the OpenSSL Licenses.
+* "OpenSSL Licenses" means the OpenSSL License and Original SSLeay License
+* under which the OpenSSL Project distributes the OpenSSL toolkit software,
+* as those licenses appear in the file LICENSE-OPENSSL.
+*/
+
+#include <stdexcept>
```
rkd-msw: Do we need this?

src/alarm_heap.cpp
```diff
@@ -0,0 +1,326 @@
...
+#include <time.h>
+#include <cmath>
+#include <arpa/inet.h>
```
rkd-msw: Do we need this?

src/alarm_heap.cpp
```diff
@@ -0,0 +1,326 @@
...
+* Project and licensed under the OpenSSL Licenses, or a work based on such
+* software and licensed under the OpenSSL Licenses.
+* "OpenSSL Licenses" means the OpenSSL License and Original SSLeay License
+* under which the OpenSSL Project distributes the OpenSSL toolkit software,
+* as those licenses appear in the file LICENSE-OPENSSL.
+*/
+
+#include <stdexcept>
+#include <time.h>
+#include <cmath>
```
rkd-msw: Do we need this?

src/alarm_heap.cpp
```diff
@@ -0,0 +1,326 @@
...
+  {
+    return AlarmSeverityChange::REDUCED_SEVERITY;
+  }
+  else
+  {
+    return AlarmSeverityChange::INCREASED_SEVERITY;
+  }
+}
+
+void AlarmInfo::set_severity(AlarmTableDef& alarm_table_def)
```
rkd-msw: Worth a comment somewhere to say that this class is the one responsible for updating alarmActiveTable behind the scenes?

rkd-msw: (If it should be - maybe it's more natural for this to just take a severity and for heap_sender to handle the AAT.)

src/alarm_heap.cpp
```diff
@@ -0,0 +1,326 @@
...
+
+void AlarmInfo::update_alarm_in_heap(AlarmDef::Severity new_severity,
+                                     uint64_t time_to_delay)
+{
+  // Update the alarm in the heap's severity, and update its time to pop
+  // (which also rebalances the heap).
+  if (new_severity != _alarm_in_heap->severity())
+  {
+    _alarm_in_heap->set_severity(new_severity);
+    _alarm_in_heap->update_pop_time(time_to_delay);
```
rkd-msw: Are we ever going to want to update the pop time but with an unchanged severity? Worth a comment to say that's impossible?

src/alarm_heap.cpp
```diff
@@ -0,0 +1,326 @@
...
+void AlarmHeap::heap_sender()
+{
+  pthread_mutex_lock(&_lock);
+
+  while (!_terminated)
+  {
+    uint64_t time_now_in_ms = Utils::get_time();
+    AlarmInHeap* alarm_in_heap = (AlarmInHeap*)_alarm_heap.get_next_timer();
+
+    while ((alarm_in_heap) &&
```
rkd-msw: Maybe clearer to do the timer-getting at the start of the while loop rather than the end:

```
AlarmInHeap* alarm_in_heap = nullptr;

while ((alarm_in_heap = (AlarmInHeap*)_alarm_heap.get_next_timer()) &&
...
```

src/alarm_heap.cpp
```diff
@@ -0,0 +1,326 @@
...
+        {
+          (*alarm).second->set_severity(alarm_table_def);
+        }
+      }
+
+      alarm_in_heap->remove_from_heap();
+      alarm_in_heap = (AlarmInHeap*)_alarm_heap.get_next_timer();
+    }
+
+    // The next alarm in the heap isn't due to pop yet. Wait until it's due.
```
rkd-msw: Duplicated comment (which doesn't make sense here, because there may be no next alarm in the heap).

src/alarm_heap.cpp
```diff
@@ -0,0 +1,326 @@
...
+    {
+      // While the next alarm in the heap is due to be sent, we:
+      //  - Pull out the alarm definition for its index and severity.
+      //  - Pass the alarm definition to the trap sender (which is responsible
+      //    for actually sending any INFORMs).
+      //  - Update the master view of the current severity for this alarm (by
+      //    updating the _all_alarms_state map).
+      //  - Finally, we then remove the alarm from the heap.
+      AlarmTableDef& alarm_table_def =
+              _alarm_table_defs->get_definition(alarm_in_heap->index(),
```
rkd-msw: Should the alarm timer just store all this context so you don't have to do these lookups?

src/alarm_heap.cpp
rkd-msw: (Or just removing this code, which simplifies things.)

src/alarm_heap.cpp
```diff
@@ -0,0 +1,326 @@
...
+    {
+      // The next alarm in the heap isn't due to pop yet. Wait until it's due.
+      struct timespec time;
+      time.tv_sec = alarm_in_heap->get_pop_time() / 1000;
+      time.tv_nsec = (alarm_in_heap->get_pop_time() % 1000) * 1000000;
+      _cond->timedwait(&time);
+    }
+    else
+    {
+      // There's no alarms in the heap. Wait until we're signalled.
```
rkd-msw: There *are* no *alarms* in the heap

src/alarm_heap.cpp
```diff
@@ -0,0 +1,326 @@
...
+
+  pthread_mutex_unlock(&_lock);
+}
+
+void AlarmHeap::update_alarm_in_heap(AlarmInfo* alarm_info,
+                                     AlarmDef::Severity severity,
+                                     uint64_t alarm_pop_time)
+{
+  _alarm_heap.insert(alarm_info->alarm_in_heap());
+  alarm_info->update_alarm_in_heap(severity, alarm_pop_time);
```
rkd-msw: Should this call `_cond->signal`, so you don't have to do so after every call?

src/alarm_heap.cpp
```diff
@@ -0,0 +1,326 @@
...
+    {
+      // There's no alarms in the heap. Wait until we're signalled.
+      _cond->wait();
+    }
+  }
+
+  pthread_mutex_unlock(&_lock);
+}
+
+void AlarmHeap::update_alarm_in_heap(AlarmInfo* alarm_info,
```
rkd-msw: I wonder if this could have a name that describes what it's doing more conceptually (`schedule_severity_change`)?

src/alarm_heap.cpp
```diff
@@ -0,0 +1,326 @@
...
+}
+
+void AlarmHeap::issue_alarm(const std::string& issuer,
+                            const std::string& identifier)
+{
+  unsigned int index;
+  unsigned int severity;
+  if (sscanf(identifier.c_str(), "%u.%u", &index, &severity) != 2)
+  {
+    TRC_ERROR("malformed alarm identifier: %s", identifier.c_str());
```
rkd-msw: Should this validation live elsewhere (at least a separate function)?

src/alarm_heap.cpp
```diff
@@ -0,0 +1,326 @@
...
+  }
+
+  AlarmTableDef& alarm_table_def = _alarm_table_defs->get_definition(index,
+                                                                     severity);
+
+  if (alarm_table_def.is_valid())
+  {
+    pthread_mutex_lock(&_lock);
+
+    AllAlarmsStateMap::iterator alarm = _all_alarms_state.find(index);
```
rkd-msw: I think a valid alarm def will always be in here, by definition.

src/alarm_heap.cpp
```diff
@@ -0,0 +1,326 @@
...
+
+    if (alarm != _all_alarms_state.end())
+    {
+      AlarmDef::Severity severity_as_enum = (AlarmDef::Severity)severity;
+      AlarmSeverityChange severity_change =
+                  (*alarm).second->new_alarm_severity(severity_as_enum);
+
+      if (severity_change == AlarmSeverityChange::INCREASED_SEVERITY)
+      {
+        TRC_DEBUG("Severity of the alarm %lu has increased", index);
```
rkd-msw: index is an int, not a long, so just %u

src/alarm_heap.cpp
rkd-msw: (looks like update_alarm_in_heap does so anyway)

src/alarm_heap.cpp
rkd-msw: Should we signal the heap in this case?

src/alarm_heap.cpp
```diff
@@ -0,0 +1,326 @@
...
+{
+  TRC_STATUS("Resyncing all alarms");
+
+  pthread_mutex_lock(&_lock);
+
+  for (AllAlarmsStateMap::iterator it = _all_alarms_state.begin();
+                               it != _all_alarms_state.end();
+                               it++)
+  {
+    AlarmDef::Severity current_severity = (*it).second->get_severity();
```
rkd-msw: Worth a comment to explain what this is doing?

src/alarm_heap.cpp
```diff
@@ -0,0 +1,326 @@
...
+}
+
+void AlarmHeap::handle_failed_alarm(AlarmTableDef& alarm_table_def)
+{
+  TRC_DEBUG("Handling a failed callback for the alarm: %lu",
+            alarm_table_def.alarm_index());
+
+  pthread_mutex_lock(&_lock);
+
+  AllAlarmsStateMap::iterator alarm =
```
rkd-msw: Again, this is probably overdefensive.

src/alarm_heap.cpp
```diff
@@ -0,0 +1,326 @@
...
+    }
+  }
+
+  _cond->signal();
+  pthread_mutex_unlock(&_lock);
+}
+
+void AlarmHeap::handle_failed_alarm(AlarmTableDef& alarm_table_def)
+{
+  TRC_DEBUG("Handling a failed callback for the alarm: %lu",
```
rkd-msw: A more specific log ("Handling an alarm which the NMS didn't respond to") might be clearer.

src/alarm_heap.cpp
rkd-msw: If you renamed `should_resend_alarm` to `is_severity_up_to_date`, would this be clearer?

include/alarm_heap.hpp
```diff
@@ -0,0 +1,198 @@
...
+  AllAlarmsStateMap _all_alarms_state;
+
+  pthread_mutex_t _lock;
```
rkd-msw: Worth a comment to say what data structures this lock protects and when you should take it?

rkd-msw:
Looks good, and the move away from singletons is good. I've been fairly thorough in my review because I think if we have a refactoring story we should get really clear, well-factored code out of it, so my standards are higher than usual.

Summarising my comments:

- it wasn't always obvious to me that the code gave the right result, particularly when resyncing and retransmitting started putting things in the heap as well as delayed clears
- I wasn't sure about the naming and organization - for example, I'm not sure I could confidently explain the difference between an AlarmInfo and an AlarmInHeap, or what it meant for an AlarmInHeap not to be in the AlarmHeap
- There were a couple of places where I thought the code was a bit too defensive - specifically, building up the _all_alarm_state map from the valid AlarmtableDefs, then double-checking that a valid AlarmTableDef had a corresponding _all_alarm_state entry. Better to say "the constructor guarantees that all valid definitions have an entry in this map" then rely on that logic.
- There are some C++ features that we've used elsewhere which I think would make the code more readable:
    - range-based for loops
    - `->` on iterator objects
    - overloading `operator<` for custom comparisons

include/alarm_heap.hpp
eleanor-merry: It can't really - I'll change the func name to make it clear that these aren't meant to be symmetrical functions

src/alarm_trap_sender.cpp
eleanor-merry: Fair enough

src/alarm_heap.cpp
eleanor-merry: No good reason! (I deleted the wrong line of code when I moved this into a function...)

PR #149, opened 2016-06-21
[Reviewer: Matt] Port the alarm agent to Centos

AlexHockey:
Hi Matt, 

Another PR, this time porting the alarm agent to centos. Again, I've tested that the RPM builds successfully, but I've not done any live testing yet. 

Thanks,
Alex.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,17 +9,24 @@ PREFIX ?= ${ROOT}/usr
 INSTALL_DIR ?= ${PREFIX}
 MODULE_DIR := ${ROOT}/modules
 
-DEB_COMPONENT := clearwater-snmp-handlers
-DEB_MAJOR_VERSION := 1.0${DEB_VERSION_QUALIFIER}
-DEB_NAMES := clearwater-snmp-handler-cdiv clearwater-snmp-alarm-agent clearwater-snmp-handler-memento-as clearwater-snmp-handler-memento clearwater-snmp-handler-astaire clearwater-snmp-alarm-agent-dbg
+PKG_COMPONENT := clearwater-snmp-handlers
+PKG_MAJOR_VERSION ?= 1.0
 
-# Add dependencies to deb-only (target will be added by build-infra)
+# Define DEB_NAMES and RPM_NAMES separately as we don't have RPMs for all the
+# components yet.
+DEB_NAMES := clearwater-snmp-handler-cdiv clearwater-snmp-alarm-agent clearwater-snmp-handler-memento-as clearwater-snmp-handler-memento clearwater-snmp-handler-astaire clearwater-snmp-alarm-agent-dbg
+RPM_NAMES := clearwater-snmp-alarm-agent
+
+# Add dependencies to deb-only and rpm-only (targets will be added by
+# build-infra). Again, these definitions are differnet as we don't have RPMs for
+# everything yet.
 deb-only: cw_alarm_agent cdiv_handler.so memento_handler.so memento_as_handler.so astaire_handler.so
+rpm-only: cw_alarm_agent
 
 INCLUDE_DIR := ${INSTALL_DIR}/include
 LIB_DIR := ${INSTALL_DIR}/lib
 
-SUBMODULES := 
+SUBMODULES :=
 
 CW_ALARM_AGENT_DIR := ${ROOT}/src
 CW_ALARM_AGENT_TEST_DIR := ${ROOT}/tests
@@ -68,9 +75,12 @@ distclean: $(patsubst %, %_distclean, ${SUBMODULES}) cw_alarm_agent_distclean
 	rm -rf ${ROOT}/build
 
 include build-infra/cw-deb.mk
-
 .PHONY: deb
 deb: build deb-only
+
+include build-infra/cw-rpm.mk
+.PHONY: rpm
+rpm: build rpm-only
 
 .PHONY: all build test clean distclean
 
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -9,17 +9,25 @@ PREFIX ?= ${ROOT}/usr
...
 
-DEB_COMPONENT := clearwater-snmp-handlers
-DEB_MAJOR_VERSION := 1.0${DEB_VERSION_QUALIFIER}
+PKG_COMPONENT := clearwater-snmp-handlers
+PKG_MAJOR_VERSION ?= 1.0
+
+# Define DEB_NAMES and RPM_NAMES separately as we don't have RPMs for all the
+# components yet.
 DEB_NAMES := clearwater-snmp-handler-cdiv clearwater-snmp-alarm-agent clearwater-snmp-handler-memento-as clearwater-snmp-handler-memento clearwater-snmp-handler-astaire clearwater-snmp-alarm-agent-dbg
+RPM_NAMES := clearwater-snmp-alarm-agent clearwater-snmp-alarm-agent-debuginfo
```
mirw: Should we changing DEB_NAMES to PKG_NAMES and make cw-rpm.mk be smart about "-dbg" -> "-debuginfo" (or automatically look for "-dbg"/"-debuginfo" versions of every package, as they're always going to be tied to the same repository as their parent)?

mirw: This would also avoid the need for the RPM_SPECS line below.

Makefile
```diff
@@ -68,9 +76,12 @@ distclean: $(patsubst %, %_distclean, ${SUBMODULES}) cw_alarm_agent_distclean
...
 include build-infra/cw-deb.mk
-
 .PHONY: deb
 deb: build deb-only
 
+include build-infra/cw-rpm.mk
+.PHONY: rpm
+rpm: build rpm-only
+
 .PHONY: all build test clean distclean
```
mirw: I think our `.PHONY` statements here are a bit broken - `all` doesn't seem to be a target, and above we have `cdiv_handler.so`, which is a target but definitely not phony.  Would you mind taking a quick look?

mirw: (Just to be clear, nothing to do with your changes.)

Makefile
AlexHockey: I didn't do this originally because it seems like it would couple rpms to deb, but on further thought I'm not sure it does. I think I'd implement this as: 

* `RPM_NAMES` defaults to `PKG_NAMES` but with the `-dbg` -> `-debuginfo` substitution applied. If you ever want to specify `RPM_NAMES` explicitly you have to specify debug packages using `-debuginfo`
* `cw-rpm.mk` spots packages ending in `-debuginfo` and doesn't try to build them. 

Sound OK?

Makefile
AlexHockey: Will do.

debian/clearwater-snmp-alarm-agent.upstart
```diff
@@ -1,37 +1,13 @@
...
-# running in different namespaces to obtain sockets in other namespaces.
-
 description "Clearwater alarm agent"
 
 start on (starting clearwater-monit)
 stop on runlevel [!2345]
 
 respawn
-# Sleep for 5 seconds on stop, to avoid respawning too quickly
-post-stop exec sleep 5
```
mirw: Why don't we need this any more?

rpm/clearwater-snmp-alarm-agent.spec
```diff
@@ -0,0 +1,31 @@
...
+systemctl enable clearwater-snmp-alarm-agent
+systemctl start clearwater-snmp-alarm-agent
+
+%preun
+systemctl stop clearwater-snmp-alarm-agent
+systemctl disable clearwater-snmp-alarm-agent
+
+%postun
+# Trigger an snmpd restart to unload the sub-agent
+systemctl stop snmpd || true
```
mirw: Do we actually need to do this?  We would for .so files, but I'm surprised for AgentX processes.

mirw:
Looks good.  Comments inline.

Makefile
mirw: As discussed, I think it would be cleaner to

*   remove `-dbg` and `-debuginfo` from `PKG_NAMES`, `DEB_NAMES` and `RPM_NAMES`
*   in cw-rpm.mk,

    *    build the spec files for all `RPM_NAMES`
    *    copy all packages named by `RPM_NAMES` or named by `RPM_NAMES` with a `-debuginfo` suffix

*   in cw-deb.mk, copy all packages named by `DEB_NAMES` or named by `DEB_NAMES` with a `-dbg` suffix
*   remove `RPM_SPECS` entirely (as it's not required).

Sound reasonable?

AlexHockey: Yep. I have implemented this but want to do some more testing before sending for re-review.

debian/clearwater-snmp-alarm-agent.upstart
AlexHockey: I deleted this and moved the code into the script block below (for consistency with other scripts). Though in retrospect this way seems better (since you sleep after the process has exited, rather than before it starts). I'll migrate the other script I saw to use this approach.

rpm/clearwater-snmp-alarm-agent.spec
AlexHockey: I'm not really sure, but:

* It's only called on post uninstall, which will rarely be executed in practice. 
* Changing this behaviour seems out of scope of porting to centos.

mirw: Fine.

Makefile
AlexHockey: It turns out that `all` is a target (defined at the top of the file) and the other targets that look like they shouldn't be phony are actually phonies.

PR #153, opened 2016-06-30
[Reviewer: Matt] Triggers rework

AlexHockey:
Hi Matt, please can you review this PR to rework the alarm agent's triggers.

The PR makes two changes: 

* Refactors the processing that is involved when the agent is triggered into a separate script which can be called directly. This is needed on centos which has a less rich triggering infrastructure. 
* Makes the triggering logic work on ubuntu and centos. 

This means that packages which want to trigger the alarm agent can either call the script directly, or use the existing debian triggering mechanism (which maps through to the script). This means we don't have to go through all our debian packages and change them immediately; we can do this as we port them to work on centos. I don't think this technical debt will cost us very much.

I've tested live with:

* A centos package that calls the trigger script directly
* A debian package that calls the trigger script directly
* A debian package that uses the built-in debian triggers mechanism


mirw:
Looks good - in particular, good compatibility thinking!

PR #157, opened 2016-07-27
[Reviewer: Ellie] Adds support for enterprise alarms MIB

AME2:
Changes here include Adding methods to the AlarmTableDef class to retrieve the new values in the above structs (that are part of the AlarmTableDef object) and tests these get parsed correctly by the JSON parser.


include/alarm_table_defs.hpp
```diff
@@ -61,12 +61,18 @@ class AlarmTableDef
...
   unsigned int alarm_index() const   {return _alarm_definition._index;} 
   AlarmDef::Cause cause() const      {return _alarm_definition._cause;}
 
-  AlarmDef::Severity severity() const    {return _severity_details._severity;}
-  const std::string& description() const {return _severity_details._description;}
-  const std::string& details() const     {return _severity_details._details;}
+  AlarmDef::Severity severity() const             {return _severity_details._severity;}
+  const std::string& description() const          {return _severity_details._description;}
+  const std::string& details() const              {return _severity_details._details;}
+  const std::string& extended_cause() const       {return _severity_details._cause;}
```
eleanor-merry: Why is this one extended_cause?

src/ut/alarm_table_defs_test.cpp
```diff
@@ -96,8 +96,37 @@ TEST_F(AlarmTableDefsTest, ValidTableDefLookup)
...
+  EXPECT_THAT(_def.effect(), StrEq("Effect"));
+  EXPECT_THAT(_def.action(), StrEq("Action"));
+  // The JSON file contains no extended details or description so we test that
+  // we have used the regular details and description in place.
+  EXPECT_THAT(_def.extended_details(), 
+              StartsWith("Monit has detected that the process has failed"));
+  EXPECT_THAT(_def.extended_description(), StrEq("Process failure"));
+}
+
+TEST_F(AlarmTableDefsTest, ExtendedFieldsDefLookup)
```
eleanor-merry: I'm not sure I see what this test is giving you beyond the cpp-common tests. Let's discuss this tomorrow.

eleanor-merry:
Changes looks good, but I've got some questions about the UTs

include/alarm_table_defs.hpp
AME2: To distinguish it from the other cause returning function. Each alarm has a cause like "SOFTWARE_ERROR" and then each severity within that alarm has an extended_cause like "The system has detected that the Sprout process has failed. Monit will automatically attempt to restart it."

I've just checked, I thought I was copying the name from the JSON parser but that actually called it detailed_cause. I can change the names for consistency.

include/alarm_table_defs.hpp
```diff
@@ -61,12 +61,18 @@ class AlarmTableDef
...
-  const std::string& description() const {return _severity_details._description;}
-  const std::string& details() const     {return _severity_details._details;}
+  const std::string& name() const     {return _alarm_definition._name;}
+  unsigned int alarm_index() const    {return _alarm_definition._index;} 
+  AlarmDef::Cause alarm_cause() const {return _alarm_definition._cause;}
+
+  AlarmDef::Severity severity() const             {return _severity_details._severity;}
+  const std::string& description() const          {return _severity_details._description;}
+  const std::string& details() const              {return _severity_details._details;}
+  const std::string& extended_cause() const       {return _severity_details._cause;}
```
eleanor-merry: Why is this still extended_cause?

src/ut/alarm_table_defs_test.cpp
eleanor-merry: I don't think you've covered the above comment in your updates yet

AME2: This test isn't giving us anything beyond the cpp-common tests. I included it because there was already a test of alarm_table_defs using RFC style traps (ValidTableDefLookup) so thought it made sense to extended this to include a similar test with Enterprise style traps. I'll delete this test.

PR #163, opened 2016-10-07
[Reviewer: Ellie] Indicates that when there are no attempts the success percent stat will report 100%

AME2:


PROJECT-CLEARWATER-MIB
```diff
@@ -3003,7 +3037,9 @@ sproutICSCFSessionEstablishmentNetworkSuccesses OBJECT-TYPE
     MAX-ACCESS  read-only
     STATUS      current
     DESCRIPTION "The number of terminating session attempts over the
-                 period that were successful from a network perspective."
+                 period that were successful. Reported in units of tens of 
+                 thousands of a percent. When there are zero attempts made 
+                 the value will be 1000000 to represent 100%.from a network perspective."
```
eleanor-merry: Some of these have the 'from a network perspective' in the wrong place

AME2: Nice one, good spot.

PR #165, opened 2016-12-14
[REVIEWER: Ellie] Add a stat to track forked INVITEs due to multiple bindings

NigelThorpe:
This is the equivalent MIB update corresponding to metaswitch/sprout#1638


NigelThorpe:
Tweaked and updated after MIB validator pointed out that I had neglected to increment a table index. :pensive:

PR #182, opened 2017-05-22
[Reviewer EM] Include alarm state at the end of the alarm OID

steelyvoid:
Hi Ellie

Can you sign off on this change for me (we discussed last week)?  I've also tweaked the MIB text to make it clear that the OID is for the alarm itself, not the alarmed object (which wouldn't have the "state" element in it, logically)

I've done a live test and can confirm that the state now comes out in the alarm message.


eleanor-merry:
:shipit:

PR #183, opened 2017-05-23
[Reviewer: MGM] Added clarification to transaction tables that 3xx is counted as failure

GAB-MS:
Assigned to @MatMeredith as the recommended reviewer - if you're not the right person, please pass it on.

It initially seems counter-intuitive that we log 3xx responses as a 'failure', given that they can lead to redirection and call success. After discussion and thinking, we decided to keep logging 3xx as a failure, but also to clarify the table descriptions to make explicitly clear that 3xx responses are logged as failures - this reduces the potential for confusion if an operator expects a significant number of 3xx responses to pass through the network.

### DIFF
--- a/PROJECT-CLEARWATER-MIB
+++ b/PROJECT-CLEARWATER-MIB
@@ -1986,7 +1986,8 @@ sproutICSCFIncomingSIPTransactionsFailures OBJECT-TYPE
     MAX-ACCESS  read-only
     STATUS      current
     DESCRIPTION "The number of failed incoming SIP transactions over the
-                 time period."
+                 time period. Note that 3xx responses are counted as 
+                 transaction failures."
     ::= { sproutICSCFIncomingSIPTransactionsEntry 5 }
 
 sproutICSCFIncomingSIPTransactionsSuccessPercent OBJECT-TYPE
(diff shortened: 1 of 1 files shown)


MatMeredith:
I can't comment on this from a spec behaviour, but if that's what you've agreed then that's fine by me.

Questions on the code changes.
- I think we should probably clarify on sproutICSCFSessionEstablishmentFailures as well?
- Does it also apply to the registration / authentication failure stats?  Can these be redirected?  (possibly not -- I just don't know).

I think you also need to update the corresponding definitions here: https://github.com/Metaswitch/stats-engine/blob/dev/stats-engine.root/usr/share/stats-engine/mibs/METASWITCH-STATS-ENGINE-MIB

GAB-MS:
You're right - it should be on the Session Establishment table as well.

A quick search suggests that 3xx responses can be given on REGISTERs - so I'll update the reg/auth stats as well.

I'm checking with CNE on how the stats engine MIB is kept up to date - I'll make sure that anything that needs to be done gets done.

MatMeredith:
I'm pretty sure the process for keeping the stats engine MIB up to date is just to update it when you change the PROJECT CLEARWATER MIB :-)

PR #187, opened 2017-06-01
[Reviewer: HJS] Add MIB automation infrastructure

GAB-MS:
Sent to @huws for internal team review - I expect to also get these changes reviewed by someone in the Clearwater team.

This adds the open-source infrastructure for generating the PROJECT-CLEARWATER-MIB, with the ability to generate the closed-source Clearwater Core MIB when it has the appropriate fragment. These changes are to make it easier for us to make additions to the MIB for all Clearwater deployments, or specifically for Project Clearwater or Clearwater Core deployments.

This adds 3 files forming the open-source infrastructure:

* CLEARWATER-MIB-COMMON: Contains the bulk of the MIB, with stats common to both types of deployment.
* CLEARWATER-MIB-PC-EXTRAS: Contains MIB entries and compliance statements specifically for Project Clearwater deployments.
* cw_mib_generator.py: Builds the PROJECT-CLEARWATER-MIB from the above files. If the CWC fragment is present in the correct place, also builds the CWC mib.

This also checks in a PROJECT-CLEARWATER-MIB built using the above infrastructure. This MIB contains all of the entries that it did previously, with the following changes:
* Separate compliance statements for common Clearwater components and Project Clearwater-specific components. (I had to use a weird OID for the latter compliance statement, so that these changes could be made without changing any stat OIDs.)
* Project Clearwater-specific Sprout stats moved into separate stat groups, but retain their original OIDs.
* Ordering changed - common Clearwater entries, followed by PC-specific entries.

The MIB passed a validation tool with no errors, as with the original PROJECT-CLEARWATER-MIB. I have also validated it by loading into snmpwalk against a Sprout node, and it looked correct.

I am expecting feedback on the script - I wasn't sure about adding error handling. I haven't added comments, as I felt it was simple and self-documenting. (I've already noticed that the copyright date is wrong - I'll change this along with any other review feedback.)


cw_mib_generator.py
```diff
@@ -0,0 +1,94 @@
...
+                       find_header(data[module_start_index+1:], 
+                                   generic_header_regex)
+
+    return module_start_index, module_end_index
+
+
+def append_compliance(data, compliance_objects):
+    compliance_regex = re.compile(r"\-\- Module Compliance")
+    compliance_section = find_section(data, compliance_regex)
+    data[compliance_section[1]:compliance_section[1]] = compliance_objects[1:]
```
huws: I don't really like the use of magic numbers, here.  To understand this, I have to work out that `find_section()` is returning line indices for the start and end of the section, and know something about why it is necessary to strip out the first line of the `compliance_objects` list.

I think it would be clearer if `find_section()` returned a simple `Section` object, that contained `start_index` and `end_index` members.

cw_mib_generator.py
```diff
@@ -0,0 +1,94 @@
...
+
+import re
+import os.path
+
+COMMON_MIB = "./CLEARWATER-MIB-COMMON"
+PC_EXTRAS = "./CLEARWATER-MIB-PC-EXTRAS"
+CWC_EXTRAS = "../CLEARWATER-MIB-CWC-EXTRAS"
+
+PC_MIB_NAME = "./PROJECT-CLEARWATER-MIB"
+CWC_MIB_NAME = "../METASWITCH-CLEARWATER-CORE-MIB"
```
GAB-MS: I still need to think about, discuss, and finalise the build process. These are easy to change once that happens.

GAB-MS:
Huw, the script has been almost entirely re-written, could you please re-review?

Things to note:

* In testing, conformance statements in the middle of a MIB are acceptable. Therefore, I've changed the script to simply append one MIB fragment to another.
* Error handling will catch all of the errors that I could think of, with useful error output in each instance. I've tested all of these errors.
* I've added an argument which allows for the CWC directory containing the CWC MIB fragment to be specified. The CWC-related file names are kept as constants, and I assume that the output would be to the same directory specified.
* I've moved the script and the PC MIB fragment into a separate folder, so that they are distinct from the auto-generated MIB.
* I've added templating to the COMMON MIB fragment. This is simply to add the title, and a comment regarding directly editing the auto-generated files.

mib-generator/cw_mib_generator.py
```diff
@@ -0,0 +1,150 @@
...
+if __name__ == "__main__":
+    args = sys.argv[1:]
+
+    # If args are supplied, try to parse the CWC MIB directory.
+    if args:
+        try:
+            arguments = docopt(__doc__, argv=args)
+        except DocoptExit:
+            sys.stderr.write(
+                "Invalid argument. Takes '--cwc-mib-dir=DIR' or no args.\n")
```
huws: Surely this isn't necessary.  Isn't the whole point of `docopt` that it will print out the usage statement for you, based on the docstring at the top of the file?  Just call `docopt()`, and let it terminate the program for you, if the parameters were wrong.

GAB-MS:
@chris-elford-metaswitch, sending over to you as discussed to allocate for review within your team.

Thanks

mib-generator/cw_mib_generator.py
```diff
@@ -0,0 +1,145 @@
...
+
+Options:
+  --cwc-mib-dir=DIR  Directory containing Clearwater Core MIB fragment
+"""
+
+import re
+import sys
+import os.path
+from string import Template
+from docopt import docopt
```
chris-elford-metaswitch: The import order isn't quite as spec'd. I think `autopep8` will fix that up for you. I'd suggest running that over all your code so you don't have to bother with PEP8 formatting yourself.

mib-generator/cw_mib_generator.py
```diff
@@ -0,0 +1,145 @@
...
+
+# Project Clearwater MIB paths
+COMMON_MIB_PATH = "./CLEARWATER-MIB-COMMON"
```
chris-elford-metaswitch: It's always bad when a script needs to be run from a specific working directory. These should all be relative to the script's location. I'd write a function to do the fiddly path manipulation. Usefully, that means you don't need to do separate `abspath` fiddling in `generate_mib` below.

mib-generator/cw_mib_generator.py
```diff
@@ -0,0 +1,145 @@
...
+
+# Statement added to top of auto-generated MIBs
+EDIT_STATEMENT = "THIS MIB IS BUILT FROM A TEMPLATE - DO NOT EDIT DIRECTLY!"
+
+
+def print_err_and_exit(error_text):
+    # Prints an error message with the specified text, and exits with code 1.
+    sys.stderr.write("ERROR: {}\nFailed to generate MIB, exiting\n"
+                     .format(error_text))
+    sys.exit(1)
```
chris-elford-metaswitch: Why not pass the error message to `sys.exit` instead of printing it out separately?

mib-generator/cw_mib_generator.py
```diff
@@ -0,0 +1,145 @@
...
+            mib_file.write(mib_contents)
+    except IOError as ioe:
+        print_err_and_exit("Could not write MIB file:\n - {}".format(ioe))
+
+
+def generate_title(mib_file_path):
+    # Try to extract MIB name from the file path, and append definitions
+    # statement.
+    try:
+        name = re.search(r"[\w\-]+$", mib_file_path).group(0)
```
chris-elford-metaswitch: This is a bit backwards. Rather than defining paths, then extracting the names from the paths, why not define the names, then write a function to generate the path?

mib-generator/cw_mib_generator.py
```diff
@@ -0,0 +1,145 @@
...
+
+    return name + " DEFINITIONS ::= BEGIN"
+
+
+def generate_mib(extras_file_path, output_mib_path):
+    # Read in the common and specific MIB fragments, fill in templated lines
+    # and write the complete MIB file.
+    full_common_mib_path = os.path.abspath(COMMON_MIB_PATH)
+    raw_common_data = read_mib_fragment(full_common_mib_path)
+    common_src_template = Template(raw_common_data)
```
chris-elford-metaswitch: Old Skool! I'm used to using template libraries that do loading and rendering for you, so you don't need all this custom machinery. For example `jinja2` was hot and fresh when I was last doing `flask` apps.

If you're into weirdness, it's worth having a look at the `Template` source code at some point. Can you convince yourself that it's a legitimate use of metaclasses?

mib-generator/cw_mib_generator.py
```diff
@@ -0,0 +1,145 @@
...
+
+    substitute_dict = { 'title_statement' : generate_title(output_mib_path),
+                        'direct_edit_statement' : EDIT_STATEMENT }
```
chris-elford-metaswitch: It's weird that these have different variable names.

mib-generator/cw_mib_generator.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+def generate_mib(extras_file_path, output_mib_path):
+    # Read in the common and specific MIB fragments, fill in templated lines
+    # and write the complete MIB file.
+    full_common_mib_path = os.path.abspath(COMMON_MIB_PATH)
+    raw_common_data = read_mib_fragment(full_common_mib_path)
+    common_src_template = Template(raw_common_data)
+
+    substitute_dict = { 'title_statement' : generate_title(output_mib_path),
```
chris-elford-metaswitch: Or `generate_path(mib_name)`?

mib-generator/cw_mib_generator.py
```diff
@@ -0,0 +1,145 @@
...
+                        'direct_edit_statement' : EDIT_STATEMENT }
+
+    try:
+        common_src = common_src_template.substitute(substitute_dict)
+    except ValueError as ve:
+        print_err_and_exit(
+            "Common MIB fragment contains an invalid placeholder:\n - {}"
+            .format(ve))
```
chris-elford-metaswitch: I'd honestly rather get a stack trace here. What's the audience for this script?

mib-generator/cw_mib_generator.py
```diff
@@ -0,0 +1,145 @@
...
+    except KeyError as ke:
+        print_err_and_exit(
+            "Common MIB fragment contains unrecognised placeholder: {}"
+            .format(ke))
+
+    extras_src = read_mib_fragment(extras_file_path)
```
chris-elford-metaswitch: It's weird to me that the different fragments are handled differently. I guess that's a symptom of `Template`'s finickiness.

mib-generator/cw_mib_generator.py
```diff
@@ -0,0 +1,145 @@
...
+    else:
+        cwc_dir = None
+        sys.stdout.write("No Clearwater Core MIB directory supplied, building "
+                         "Project Clearwater MIB only\n")
+
+    full_pc_output_mib_path = os.path.abspath(PC_MIB_OUTPUT_PATH)
+    full_pc_fragment_path = os.path.abspath(PC_EXTRAS_PATH)
+
+    sys.stdout.write("Generating Project Clearwater MIB at: '{}'\n"
+                     .format(full_pc_output_mib_path))
```
chris-elford-metaswitch: `print` is more idiomatic.

mib-generator/cw_mib_generator.py
```diff
@@ -0,0 +1,145 @@
...
+            .format(ke))
+
+    extras_src = read_mib_fragment(extras_file_path)
+
+    full_mib_data = common_src + extras_src
+    write_mib_file(output_mib_path, full_mib_data)
+
+
+if __name__ == "__main__":
+    args = sys.argv[1:]
```
chris-elford-metaswitch: You should never need to access `sys.argv` - it's a sign that you aren't using your higher-level tools properly.

mib-generator/cw_mib_generator.py
```diff
@@ -0,0 +1,145 @@
...
+
+    # If args are supplied, try to parse the CWC MIB directory.
+    if args:
+        arguments = docopt(__doc__, argv=args)
```
chris-elford-metaswitch: This is weird.
* It's not the standard idiom for using `docopt`.
* You've got two seemingly identical variables, `args` and `arguments` - that's always a hint you're doing something wrong.
* You don't want to check whether you've got args - that's inextensible, and doesn't capture the meaning of what you are trying to test. You want to check whether you have a `--cwc-mib-dir`.

So, you want:
```python
args = docopt(__doc__)  # Standard docopt invocation.
if args['--cwc-mib-dir']:
```
If that doesn't work (it's a while since I've used `docopt`), try using `args.get('--cwc-mib-dir', None)`

mib-generator/cw_mib_generator.py
```diff
@@ -0,0 +1,145 @@
...
+
+        # Validate the parsed directory.
+        cwc_dir = os.path.abspath(arguments['--cwc-mib-dir'])
+        full_cwc_fragment_path = os.path.join(cwc_dir, CWC_EXTRAS_NAME)
+        full_cwc_output_mib_path = os.path.join(cwc_dir, CWC_MIB_OUTPUT_NAME)
+
+        if not os.path.isfile(full_cwc_fragment_path):
```
chris-elford-metaswitch: This is a little unidiomatic. In Python, we usually ask permission rather than forgiveness. Try to do what you want to do, then catch the exception if it goes wrong.
* Exceptions are relatively cheap in Python
* Your script isn't going to do anything nasty if it does hit an exception.

mib-generator/cw_mib_generator.py
```diff
@@ -0,0 +1,145 @@
...
+
+    full_pc_output_mib_path = os.path.abspath(PC_MIB_OUTPUT_PATH)
+    full_pc_fragment_path = os.path.abspath(PC_EXTRAS_PATH)
+
+    sys.stdout.write("Generating Project Clearwater MIB at: '{}'\n"
+                     .format(full_pc_output_mib_path))
+
+    generate_mib(full_pc_fragment_path, full_pc_output_mib_path)
+
+    sys.stdout.write("Successfully generated Project Clearwater MIB!\n")
```
chris-elford-metaswitch: These `print` statements make the code a little hard to re-use as there's no (non-hacky) way of suppressing them. I'd usually use logging for this sort of thing. You're probably fine, but worth bearing in mind.

mib-generator/cw_mib_generator.py
```diff
@@ -0,0 +1,145 @@
...
+        sys.stdout.write("No Clearwater Core MIB directory supplied, building "
+                         "Project Clearwater MIB only\n")
+
+    full_pc_output_mib_path = os.path.abspath(PC_MIB_OUTPUT_PATH)
+    full_pc_fragment_path = os.path.abspath(PC_EXTRAS_PATH)
+
+    sys.stdout.write("Generating Project Clearwater MIB at: '{}'\n"
+                     .format(full_pc_output_mib_path))
+
+    generate_mib(full_pc_fragment_path, full_pc_output_mib_path)
```
chris-elford-metaswitch: I'd pull this into a `main()` function and move it to the top of the file. That way:
* The script doesn't read backwards.
* It's easy to write entry points for a `setup.py` package in the future if you decide you want to install your code as a package.

Not important, but a good habit to form.

mib-generator/cw_mib_generator.py
GAB-MS: I wasn't aware that you could do so - I'd only ever seen `sys.exit` used with an integer exit code passed to it. Looking at the docs, I can't see any reason _not_ to just pass the error message in.

mib-generator/cw_mib_generator.py
GAB-MS: So my justification here was that the full stack trace wasn't too helpful - the bit of information that we append here tells the user where the error is, and I felt that adding the extra line adds clarification to an error which was not immediately obvious if you aren't familiar with `Template`.

Ultimately, this might be a sign that you're right up above and that I should be looking into a different templating library. I went with `Template` for simplicity, as it seemed easy and clear - but I'll look into `jinja2` as suggested.

mib-generator/cw_mib_generator.py
GAB-MS: Because there's no templating going on in the 'additional' fragments, only in the core fragment (specifically for the title statement, and to add a statement about direct editing).

mib-generator/cw_mib_generator.py
chris-elford-metaswitch: `jinja2` is great, but it's probably overkill here. Sounds like you're justified in what you are doing, so keep doing it.

mib-generator/cw_mib_generator.py
chris-elford-metaswitch: Is that right?

mib-generator/cw_mib_generator.py
GAB-MS: I assume you mean 'ask forgiveness rather than permission'!

I did originally want to write this using a `try`/`except` block, but I couldn't find a nice way to do so. I want to check at this stage whether the supplied file exists, rather than hitting the error when we try to read it later on, and I felt that this looked nicer than opening the file solely to check that it's readable.

I'm open to thoughts on changing the design here, though.

mib-generator/cw_mib_generator.py
GAB-MS: I believe so. I wanted to keep it simple by just appending the additional entries to the core entries - but this wasn't easy to do with the title statement. We also wanted to add a warning statement about direct edits, which wouldn't appear on the fragments themselves!

Thinking about it, I should combine the two templated variables above into a single variable, as they're right next to each other.

mib-generator/cw_mib_generator.py
GAB-MS: Would you apply the ordering logic to the other functions as well? i.e. after `main()` at the top, would you then have `generate_mib`, then the functions it depends on, and so on?

mib-generator/cw_mib_generator.py
chris-elford-metaswitch: What I'm questioning is why you

> want to check at this stage whether the supplied file exists, rather than hitting the error when we try to read it later on

Does that actually improve your code/design?

mib-generator/cw_mib_generator.py
chris-elford-metaswitch: Not sure. I tend to follow the guideline that source files should start at a high level, then introduce lower level concepts as you read on. That way it's easy to pick up the high-level structure of the code, and you only need to look at the low-level detail you are interested in.

mib-generator/cw_mib_generator.py
GAB-MS: I think I should move the CWC MIB to be generated first, so that we can't generate a PC MIB and _then_ hit this issue, which is why I added this check in the first place. If the CWC MIB is generated first, this check is unnecessary, as you're pointing out.

mib-generator/cw_mib_generator.py
GAB-MS: I've just tried `autopep8` against the script, and the only thing it diff's is a bit of whitepace in `substitute_dict`. No complaints about the import statements. This seems to go against what you expected?

GAB-MS:
I've pushed updates to address your review comments. @chris-elford-metaswitch, could you please re-review?

With regards to the script itself, note:
* Importing `print_function` allows the script to run on both Python 2.7 and >3.3
* Script can now be run from anywhere; the hard-coded paths have been changed to hard-coded file names
* I've changed all of the exception blocks to re-raise the exception for a full stack trace, adding an extra bit of error text beforehand for clarification of common errors
* I've continued with using `Template` for simplicity, but have reduced it to a single templated variable
* When checking whether `smilint` is successful, the script checks for any `stderr` output; this is necessary as `smilint` exits with code 0 if it _runs_ successfully but finds errors
* The call to `smilint` only verifies up to severity level 3 (see [here](https://www.ibr.cs.tu-bs.de/projects/libsmi/smilint.html) for a description of severities). This covers all errors, but will ignore non-error warnings. Should this be increased to cover all warnings?

Running `smilint` required the addition of an `smilint.conf` file in `mib-generator` to include the path to the required IETF MIBs; it currently assumes that they will be located at `/var/lib/mibs/ietf`.

I've deleted the checked in `PROJECT-CLEARWATER-MIB`. The current plan is to run this script as part of an existing Jenkins job, or to create a new job for it, which should also upload the MIB to our repo or somewhere else appropriate.

mib-generator/cw_mib_generator.py
```diff
@@ -0,0 +1,134 @@
...
+    try:
+        cmd = ['smilint', '-s', '-c', smilint_conf, mib_path]
+        p = subprocess.Popen(cmd, stderr=subprocess.PIPE)
+        _, stderr = p.communicate()
+
+    except:
+        print("ERROR: Could not run smilint!")
+        raise
+
+    if p.returncode != 0 or len(stderr) != 0:
```
chris-elford-metaswitch: `stderr` is more idiomatic than `len(stderr) != 0`

GAB-MS:
Alright, third time lucky? I've pushed up changes to address most of your concerns - please take a look over and make sure you're happy with how it's documented, today if possible.

The one concern I didn't address is "Why do you need custom functions to read and write files?" I felt that it made the script easier to read to break them off into separate functions, mostly to keep the size of `generate_mib` down, and it improves future reusability. That's the case with all of the functions - they could have all been written in a single giant `generate_mib` function. Please push back if you disagree with this splitting though.

As for CI - the plan is to either integrate this script into one of the existing build jobs on arcturus, or (more likely) create a new job that runs it when any MIB fragment is changed. The resulting MIBs won't be checked in to a code repo, but they will be stored in a package repo. But I'll discuss this with people later today to form a more concrete plan - this won't get merged until that plan is complete, and I'll make sure you're happy with it too.

GAB-MS:
Bah. And that's after telling myself to make sure I don't miss the TODO. Thanks for spotting that...

GAB-MS:
Okay, I've added and verified Makefile changes that include the auto-generated MIB. When clearwater-snmp-handlers is built, it will now build the MIB and include it in the clearwater-snmp-alarm-agent package. I verified this on the package uploaded to my repo:

```shell
$ dpkg -c clearwater-snmp-alarm-agent_1.0-170609.104135_amd64.deb | grep MIB
-rw-r--r-- root/root    248440 2017-06-09 09:41 ./usr/share/clearwater/mibs/PROJECT-CLEARWATER-MIB

```

Going to merge this in, and send an email around Clearwater about the change.

PR #189, opened 2017-06-09
[Reviewer: CNE] Automated mib build infra fix

GAB-MS:
Fix to my recent PR. I was using the default Python environment for the script, which caused issues on a build system where docopt was not installed in the default environment.

This PR modifies the Makefile to create a virtualenv for the mib generator. Instead of using a `_setup.py` script for setting up the environment, like most of our Python packages, I felt it more appropriate to go with the simpler approach here - those setup scripts are aimed at installable Python packages, rather than a script executed at build time, and the only non-default package we need is docopt. I also only included `envclean` and not `pyclean`, as we don't generate any eggs, compiled Python code, or Python coverage reports.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,7 +18,7 @@
 debian/copyright
 build
 cw_alarm_agent
-/_env/
+_env/
 
 # FV tests
 fvtest-snmpd.out
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@
 debian/copyright
 build
 cw_alarm_agent
+/_env/
 
 # FV tests
 fvtest-snmpd.out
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -18,6 +18,7 @@
 debian/copyright
 build
 cw_alarm_agent
+/_env/
```
chris-elford-metaswitch: You probably don't want to '/' here.

Makefile
```diff
@@ -61,6 +63,12 @@ astaire_handler.so:
 	${MAKE} -C ${CW_ALARM_AGENT_DIR} $@
 
 .PHONY: cw_alarm_agent CW_ALARM_AGENT_test cw_alarm_agent_clean cw_alarm_agent_distclean cw_mib cdiv_handler.so memento_handler.so memento_as_handler.so astaire_handler.so
+env: $(ENV_DIR)/bin/python
```
chris-elford-metaswitch: You could use a blank line here.

PR #191, opened 2017-06-12
[Reviewer: CNE] Add output directory argument to MIB generator

GAB-MS:
@chris-elford-metaswitch could you review today if possible? If not, please reassign and let me know.

Small addition to cw_mib_generator.py to add argument specifying MIB output directory, as suggested by @richardwhiuk. Amends both script, and the Makefile to supply the output directory. This simplifies our build processes as it reduces the amount of copying around; we generate the MIB(s) directly where we want them.

In addition:
* I've moved the PC MIB generation into the if/else block for the `cwc-mib-dir` argument. Logic behind this is that I couldn't think of a scenario where we'd want to generate both at once. Please push back if you disagree here.
* Bunch of variable names changed to add clarity.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,7 @@ cw_alarm_agent_clean:
 cw_alarm_agent_distclean: CW_ALARM_AGENT_clean
 
 cw_mib: env
-	${ENV_DIR}/bin/python ${ROOT}/mib-generator/cw_mib_generator.py && mv ${ROOT}/mib-generator/PROJECT-CLEARWATER-MIB ${ROOT}/clearwater-snmp-alarm-agent.root/usr/share/clearwater/mibs/
+	${ENV_DIR}/bin/python ${ROOT}/mib-generator/cw_mib_generator.py ${ROOT}/clearwater-snmp-alarm-agent.root/usr/share/clearwater/mibs/
 
 cdiv_handler.so:
 	${MAKE} -C ${CW_ALARM_AGENT_DIR} $@
--- a/mib-generator/cw_mib_generator.py
+++ b/mib-generator/cw_mib_generator.py
@@ -64,7 +64,7 @@ def main(args):
     if args['--cwc-mib-dir']:
         cwc_dir = os.path.abspath(args['--cwc-mib-dir'])
 
-        print("Generating Clearwater Core MIB at {}".format(cwc_dir))
+        print("Generating Clearwater Core MIB at {}".format(output_dir))
         generate_mib(common_dir,
                      cwc_dir,
                      output_dir,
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -48,7 +48,7 @@ cw_alarm_agent_clean:
 cw_alarm_agent_distclean: CW_ALARM_AGENT_clean
 
 cw_mib: env
-	${ENV_DIR}/bin/python ${ROOT}/mib-generator/cw_mib_generator.py && mv ${ROOT}/mib-generator/PROJECT-CLEARWATER-MIB ${ROOT}/clearwater-snmp-alarm-agent.root/usr/share/clearwater/mibs/
+	${ENV_DIR}/bin/python ${ROOT}/mib-generator/cw_mib_generator.py ${ROOT}/clearwater-snmp-alarm-agent.root/usr/share/clearwater/mibs/
```
chris-elford-metaswitch: You need quote marks to cope with a ROOT directory with spaces in its name.

mib-generator/cw_mib_generator.py
```diff
@@ -84,23 +99,24 @@ def generate_mib(common_dir, extras_dir, common_mib, extras_mib, output_mib):
...
         extras_dir (str): Directory containing extra MIB fragment.
-        common_mib (str): File name of common MIB.
-        extras_mib (str): File name of extras MIB.
-        output_mib (str): File name for writing output MIB. NOTE: This will
-            over-write any existing file with the given name in extras_dir!
+        output_dir (str): Output directory for full MIB.
+        common_mib_name (str): File name of common MIB.
+        extras_mib_name (str): File name of extras MIB.
+        output_mib_name (str): File name for full MIB. NOTE: If this file
+            already exists in output_dir, it will be over-written!
```
chris-elford-metaswitch: It's all very well noting that here, but you haven't noted it in the user documentation where it might save someone from making a mistake.

Makefile
GAB-MS: Would this apply to all of the variable calls in the Makefile? This isn't the only place that `${ROOT}` is called, and you have other variables such as `ENV_DIR`, `MAKE` and `SUBMODULES`. Should these all be in quotes?

mib-generator/cw_mib_generator.py
GAB-MS: Good point - I'll add it to the usage doc at the top.

Makefile
chris-elford-metaswitch: It's not worth it is it?

GAB-MS: Sorry, I'm not quite clear - are you suggesting that we shouldn't make this change?

I'm happy to do so, but I don't think it will help much - a quick browse of our other Makefiles shows variables without quote marks being used in lots of them. I suspect a file structure containing spaces will break most of our builds.

chris-elford-metaswitch: Exactly - we do this everywhere, so fixing it here doesn't make any difference. It's not causing us any problems right now, so it shouldn't get in the way of your merge.

PR #196, opened 2017-07-25
[Reviewer: Rob] Add command to show active alarms

richardwhiuk:
As discussed - I think you've already tested this live on a node, so we just need a command to invoke it.

### DIFF
--- a/clearwater-snmp-alarm-agent.root/usr/share/clearwater/bin/show_active_alarms
+++ b/clearwater-snmp-alarm-agent.root/usr/share/clearwater/bin/show_active_alarms
@@ -1,3 +1,14 @@
 #!/bin/bash
 
+# @file show_active_alarms
+#
+# Query SNMP for the currently active Clearwater alarms
+#
+# Copyright (C) Metaswitch Networks 2017
+# If license terms are provided to you in a COPYING file in the root directory
+# of the source code repository by which you are accessing this code, then
+# the license outlined in that COPYING file applies to your use.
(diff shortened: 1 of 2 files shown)


clearwater-snmp-alarm-agent.root/usr/share/clearwater/bin/show_active_alarms
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
```
eleanor-merry: Can you add a copyright statement?

richardwhiuk: Done

rkd-msw:
Looks good.

For the record, I'm not expecting this to be exposed in the customer documentation - it's just a convenience wrapper for an snmpwalk command that's otherwise hard to figure out, and having it makes us more likely to spot alarms bugs.

Example output when there are alarms:

```
[spn]clearwater@sprintlab-kc-spn-2:~$ snmpwalk -v2c -c clearwater localhost -Ov -OQ .1.3.6.1.2.1.118.1.2.2.1.11
"Sprout: Communication error to session continued Application Server."
```

and when there aren't (which isn't very user-friendly):

```
[scn]clearwater@sprintlab-kc-scn-3:~$ snmpwalk -v2c -c clearwater localhost -Ov -OQ .1.3.6.1.2.1.118.1.2.2.1.11
No Such Instance currently exists at this OID
```

Customers should still monitor alarms with an NMS, not this.

PR #205, opened 2017-10-06
[Reviewer RJW2] Ensure wrapper death doesn't result in duplicate alarm agent processes

steelyvoid:
Hi Richard

Can you check that you're happy with this upstart change (and another in clearwater-infrastructure which I'll link).  This stems from a scary instance I've seen of an OOM killer wiping out the top level alarm agent wrapper process, with the result that upstart brought up a new copy of it, even though the cw_alarm_agent process (a grandchild of the original wrapper process) was still running, so we ended up with two alarm agents running.

I've tested that upstart operations (restart, stop, start etc) work as before, with the bonus that if the wrapper process is killed manually, we now get rid of the grandchild.

Note that I think the clearwater-infrastructure instances are the only other ones I could come up with.  The other clearwater-* upstart jobs exec their targets directly (rather than via wrapper scripts), so are not subject to this grandchilding,

### DIFF
--- a/debian/clearwater-snmp-alarm-agent.upstart
+++ b/debian/clearwater-snmp-alarm-agent.upstart
@@ -10,10 +10,3 @@ post-stop exec sleep 5
 script
   /usr/share/clearwater/bin/clearwater-snmp-alarm-agent-wrapper
 end script
-
-# Be certain that the alarm agent itself is no longer running.
-# Otherwise, if the wrapper script is killed, the alarm agent
-# won't be, and we'll end up with two copies of it.
-post-stop script
-  pkill -9 -f -x '/usr/share/clearwater/bin/cw_alarm_agent.*' || true
-end script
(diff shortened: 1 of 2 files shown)


steelyvoid:
Related to https://github.com/Metaswitch/clearwater-infrastructure/pull/497

debian/clearwater-snmp-alarm-agent.upstart
```diff
@@ -10,3 +10,10 @@ post-stop exec sleep 5
 script
   /usr/share/clearwater/bin/clearwater-snmp-alarm-agent-wrapper
 end script
+
+# Be certain that the alarm agent itself is no longer running.
+# Otherwise, if the wrapper script is killed, the alarm agent
+# won't be, and we'll end up with two copies of it.
+post-stop script
+  pkill -9 -f -x '/usr/share/clearwater/bin/cw_alarm_agent.*' || true
```
richardwhiuk: Would we better to do this before attempting to spawn the next one, rather than doing it as we exit? That feels better from a reliability perspective?

richardwhiuk: Also, should we actually be doing this in the wrapper script, rather than doing it in the upstart job?

steelyvoid:
Thanks Richard. 

Does this rework look better?

PR #207, opened 2017-11-13
[Reviewer: KH1] Add table for worker thread queue success/failure

sathiyan-sivathas:
Companion PR to Metaswitch/sprout#2010


sathiyan-sivathas:
I've tested by querying the new table while the live tests were running and the output looked sensible so I have some level of confidence that this is right.

PR #209, opened 2017-11-30
[Reviewer: Adam] Add clearwater-snmp-alarm-agent-dbg to Makefile, so it gets included in repo

mirw:
Adam,

Just noticed the debug package was missing from `Makefile`, and hence from our repo.  Fix is trivial, and safe.

Having said that, we should hold off merging this until branches are cut next week.

Matt


johadalin:
Yup. Can't see this doing us any harm. I'll leave it with you to merge post-cut

PR #210, opened 2017-12-01
[Reviewer: Richard] Reuse SNMP Agent's lock for alarm scheduler, to close memory-leaking …

mirw:
…window condition

Please can review?  UT and live-tested (with helgrind) - still waiting to confirm the memory leak is actually closed.

### DIFF
--- a/src/alarm_scheduler.cpp
+++ b/src/alarm_scheduler.cpp
@@ -43,7 +43,7 @@ void SingleAlarmManager::change_schedule(AlarmDef::Severity new_severity,
 AlarmScheduler::AlarmScheduler(AlarmTableDefs* alarm_table_defs, 
                                std::set<NotificationType> snmp_notifications,
                                std::string hostname,
-                               pthread_mutex_t* lock) :
+                               pthread_mutex_t& lock) :
   _terminated(false),
   _alarm_table_defs(alarm_table_defs),
   _lock(lock)
@@ -52,9 +52,9 @@ AlarmScheduler::AlarmScheduler(AlarmTableDefs* alarm_table_defs,
 
   // Create the condition variables.
 #ifdef UNIT_TEST
-  _cond = new MockPThreadCondVar(_lock);
+  _cond = new MockPThreadCondVar(&_lock);
 #else
-  _cond = new CondVar(_lock);
+  _cond = new CondVar(&_lock);
 #endif
 
   // Populate the alarm state map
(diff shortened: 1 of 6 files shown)


mirw:
Fixes https://github.com/Metaswitch/clearwater-issues/issues/2655.

mirw:
As discussed, I've switched from a `pthread_mutex_t*` to a `pthread_mutex_t&`.  Passing back for re-review.

src/alarm_scheduler.cpp
```diff
@@ -298,7 +299,8 @@ void AlarmScheduler::handle_failed_alarm(AlarmTableDef& alarm_table_def)
   TRC_DEBUG("Handling an alarm (%u) the NMS didn't respond to",
             alarm_table_def.alarm_index());
 
-  pthread_mutex_lock(&_lock);
+  // No need to take a lock - this is only called from Net-SNMP, which must
+  // already have been holding it.
```
richardwhiuk: Should we use a recursive lock, which would negate this? If not, we should also add a function comment that this must only be called when you are already holding the lock, and note that NetSNMP satisfies this condition.

src/alarms_agent.cpp
```diff
@@ -172,15 +172,15 @@ int main (int argc, char **argv)
...
-  AlarmScheduler* alarm_scheduler = new AlarmScheduler(alarm_table_defs, snmp_notifications, hostname);
-  AlarmReqListener* alarm_req_listener = new AlarmReqListener(alarm_scheduler);
-
   init_alarmModelTable(*alarm_table_defs);
   init_ituAlarmTable(*alarm_table_defs);
   init_alarmActiveTable(local_ip);
  
   init_snmp_handler_threads("clearwater-alarms");
 
+  AlarmScheduler* alarm_scheduler = new AlarmScheduler(alarm_table_defs, snmp_notifications, hostname, SNMP::Agent::instance()->get_lock());
```
richardwhiuk: Is this ordering change pertinent? If so, can you add a comment?

src/ut/alarm_scheduler_test.cpp
```diff
@@ -472,9 +473,9 @@ TEST_F(AlarmSchedulerTest, AlarmFlicker)
   }
 
   cwtest_advance_time_ms(AlarmScheduler::ALARM_REDUCED_DELAY);
-  pthread_mutex_lock(&(_alarm_scheduler->_lock));
+  pthread_mutex_lock(&_alarm_scheduler->_lock);
```
richardwhiuk: Could use `_lock` here (and the other similar places) I think. Delving into the internals of a class to grab a lock is a bit icky, but locking on something you own, and passed a reference to the class under test is cleaner.

src/alarm_scheduler.cpp
mirw: Happy to add a function comment.  I'm a bit nervous about recursive locks in the SNMPAgent, as it is absolutely critical for performance in its other uses (e.g. in Sprout).

src/alarms_agent.cpp
mirw: Yes.  I've added a comment.

src/ut/alarm_scheduler_test.cpp
mirw: The indirection through `_alarm_scheduler` isn't part of my change, and (since I've already live-tested) I want to draw the line here and leave the code as-is.  OK?

mirw: Done.

PR #214, opened 2018-01-26
[Reviewer: Adam] File descriptor MIB addition

kenferguson:
This is the ASN.1 definition - I've run it through an online validator and it seems to be syntactically correct.


METASWITCH-OS-MIB
```diff
@@ -0,0 +1,74 @@
...
+    MAX-ACCESS  read-only
+    STATUS      current
+    DESCRIPTION "Current number of file descriptors in use."
+    ::= { os 2 }
+
+osFileDescriptorsPerc OBJECT-TYPE
+    SYNTAX      Unsigned32
+    MAX-ACCESS  read-only
+    STATUS      current
+    DESCRIPTION "Current percentage of file dessciptors in user."
```
johadalin: de**ss**criptors in use**r**

PR #216, opened 2018-01-29
Fix up Metaswitch OS MIB

kenferguson:
This fixes the bugs we discussed with the ASN.1 file where the tree had an extra element and had some invalid references.

Live tested querying the three entries by name.

Note that I've changed to Integer32. I can't used unsigned integer because the bash handler option doesn't support that, only integer, gauge, counter, timeticks, ipaddress, objectid, or string.

### DIFF
--- a/METASWITCH-OS-MIB
+++ b/METASWITCH-OS-MIB
@@ -18,9 +18,9 @@ IMPORTS
            Tel: +44 20 8366 1177"
         DESCRIPTION  "This MIB module defines OS statistics MIBs"
 
-        ::= { dataConnection 9 }
+        ::= { dataConnection 17 }
 
--- The object identifier subtree for the ProjectClearwater MIBs.
+-- The object identifier subtree for the Metaswitch OS MIBs.
 
 memberBody      OBJECT IDENTIFIER ::= { iso 2 }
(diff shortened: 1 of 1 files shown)


johadalin:
Looks good. I've just  double checked, and it seems to work as expected. Used the following commands:
```
[spn]clearwater@kaf-clearwater-site1-spn-1:~$ snmpgetnext -v2c -c clearwater localhost METASWITCH-OS-MIB::osFileDescriptorsCurrent
METASWITCH-OS-MIB::osFileDescriptorsCurrent.0 = INTEGER: 1824
[spn]clearwater@kaf-clearwater-site1-spn-1:~$ snmpgetnext -v2c -c clearwater localhost METASWITCH-OS-MIB::osFileDescriptorsMax
METASWITCH-OS-MIB::osFileDescriptorsMax.0 = INTEGER: 1049600
[spn]clearwater@kaf-clearwater-site1-spn-1:~$ snmpgetnext -v2c -c clearwater localhost METASWITCH-OS-MIB::osFileDescriptorsPercent
METASWITCH-OS-MIB::osFileDescriptorsPercent.0 = INTEGER: 0
```

Happy that this resolves the issues we were hitting :)

PR #217, opened 2018-01-30
[Reviewer: Adam] Include MIBs in debian packages

kenferguson:
This change is to allow me to include the new MIB in the debian package. Two changes:
- move the MIB file to the relevant directory
- un-gitignore that directory and replace with individual files. 

Happy to take suggestions if you think there's a better way of achieving that goal - I've held off testing in case you feel there's a better approach.


johadalin:
So the two you've listed are the ones put there by the autogen? 
Don't know if it's possible in a gitignore, but might be worth a comment to that effect.

Otherwise, this seems sensible, and a good way to get them  onto the node. If i can get my test rig running, we can test the install on those nodes

johadalin:
See how that goes, and can then merge if it works

kenferguson:
Yep - comments are allowed and now added. Built the mibs and the autogenerated files are still successfully ignored.

